PR #21, opened 2018-07-06
Adding regression example for Bike Sharing Demand

NMZivkovic:


samples/examples/Reggression_BikeSharingDemands/BikeSharingDemand/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            var fastTreeTweedieModel = new ModelBuilder(trainingDataLocation, new FastTreeTweedieRegressor()).BuildAndTrain();
+            var fastTreeTweedieMetrics = modelEvaluator.Evaluate(fastTreeTweedieModel, testDataLocation);
+            PrintMetrics("Fast Tree Tweedie", fastTreeTweedieMetrics);
+
+            var additiveModel = new ModelBuilder(trainingDataLocation, new GeneralizedAdditiveModelRegressor()).BuildAndTrain();
+            var additiveMetrics = modelEvaluator.Evaluate(additiveModel, testDataLocation);
+            PrintMetrics("Generalized Additive Model", additiveMetrics);
+
+            var stohasticDualCorordinateAscentModel = new ModelBuilder(trainingDataLocation, new StochasticDualCoordinateAscentRegressor()).BuildAndTrain();
```
CESARDELATORRE: Please, fix. Change from "stohastic" to "stochastic".

samples/examples/Reggression_BikeSharingDemands/BikeSharingDemand/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+            var stohasticDualCorordinateAscentMetrics = modelEvaluator.Evaluate(stohasticDualCorordinateAscentModel, testDataLocation);
```
CESARDELATORRE: Please, fix. Change from "stohastic" to "stochastic".

samples/examples/Reggression_BikeSharingDemands/BikeSharingDemand/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+            PrintMetrics("Stochastic Dual Coordinate Ascent", stohasticDualCorordinateAscentMetrics);
```
CESARDELATORRE: Please, fix. Change from "stohastic" to "stochastic".

CESARDELATORRE:
@NMZivkovic - Thanks for this PR with a sample app. I've been reviewing it and it is pretty neat/clean code, re-using code as much as possible, then training and evaluating with multiple learners.
A minor typo that I found:
There are a few objects/variables named with "**stohastic**" instead of "**stochastic**". It is just missing a 'c'.

Can you fix that?

Also, be aware that in the next minor release (v0.4) we're going to release significant breaking changes in the API. Are you ok if we merge this PR and you'll need to fix it because the upcoming breaking changes?

Again, thanks for the example, I find it pretty neat with clean code where you invested your time. 👍

samples/examples/Reggression_BikeSharingDemands/BikeSharingDemand/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            VisualizeTenPredictionsForTheModel(fastTreeTweedieModel, testDataLocation);
+            fastTreeTweedieModel.WriteAsync(@".\Model.zip");
```
CESARDELATORRE: I guess you are generating the .ZIP model file for the FastTreeTweedie learner because it has the best "R2 Score == 0.92". 
As a "nice to have", it would be nice to either generate one .ZIP file per approach/learner with a different name (just like the "ML.NET Model Builder UI app" does when doing multiple trainings) or if you generate a single model .ZIP file, it would be good to instead of doing it "hard-coded", you'd compare the metrics and decide to generate the model based on that comparison and generating the best one, dynamically, not hard-coded for the FastTreeTweedie learner. - This is just an idea/improvement, not really needed. 👍

NMZivkovic:
Hello @CESARDELATORRE ,

Thank you for the review. Glad that everything is fine!
I will fixed the typo and I don't mind changes in the API, it will be handled when 0.4 is out.
Also, then I will introduce this improvement, it is a really nice idea.
Thanks again!

CESARDELATORRE:
Hey! - I just realized of another important improvement.
Since you encapsulated the training code within the **ModelBuilder** class, you are executing the data preparation for every different algorithm/Learner.
The data preparation in the pipeline should be done just once because it is the same data for every different Learner.
The improvement should be to extract the data preparation so it runs just once, then make multiple calls or multiple training, one per learner. 
I know that your data prep is simple, but the way you have the code implemented could derive in a pattern where other data transforms would be executed one time per training, that's wrong. Data transforms can be done just once, ok? 👍

CESARDELATORRE:
Hey! - Sorry, I realized this is going to be possible with the upcoming new API but not with the current "pipeline" API. I'll approve/merge this PR and when migrating the example to the new API, this should be improved, ok? 👍

CESARDELATORRE:
@NMZivkovic - One question. This sample (BikeSharingDemand) is using two DataSets that we made available here:
https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/examples/Regression_BikeSharingDemand/BikeSharingDemand/Data

The question is, where is this data coming from? Did you get it from any public DataSet that we should reference to? 

Thanks,
Cesar

NMZivkovic:
@CESARDELATORRE 
Hi Cesar,

Yes, it is the public dataset. You can find it here - https://archive.ics.uci.edu/ml/datasets/bike+sharing+dataset

Regards,
Nikola

CESARDELATORRE:
Great, I'll update that info in a README page for the sample.
If you'd like to include any intro to the sample, add it to the README, ok?

PR #26, opened 2018-07-26
Added feature: Paint regression distribution chart/image

CESARDELATORRE:
I basically added the following feature which evaluates the model and paints a regression distribution chart. It will show up right after training and evaluating the model with 100 cases (open to configuration) coming from a test file:

image

The image can be generated as a .png file (show by Microsoft Photos or the default app for graphics) or as a .svg file (shown by the default browser).

I think we should implement a similar approach for most of the samples which just have a console application and a training process. It is illustrative about the model's accuracy.
It will be especially interesting for clustering-based models.


CESARDELATORRE:
@GalOshri and @codemzs - Please, take a look to it before I merge the PR, ok?
The chart/diagram will pop up for the user right at the end of the process, after training and trying 100 sample tests that are also represented in the regression chart.

PR #32, opened 2018-08-01
[WIP] Add F# samples and move C# samples to "csharp"

dsyme:
This adds the F# getting-started samples.  

I've only compiled so far, against my on-machine preview of ML.NET

Things to do:
* [ ] Depends on ML.NET 0.4.0 with https://github.com/dotnet/machinelearning/pull/600 and https://github.com/dotnet/machinelearning/pull/616
* [ ] Make sure the samples run 
* [ ] Consider using XPlot as it is more normal for F#
* [ ] Consider making available as F# scripts rather than just compiled code (these would run on .NET Framework).  Adding a pivot "compiled" and "scripting" may make senses, I'm not sure.
* [ ] Some "force modules to load" additions may be needed, I'm not sure
* [ ] Check the addition of the "csharp" pivot hasn't screwed anything up
* [ ] remove stuff that shouldn't be here like the change to nuget.config


dsyme:
nb. this won't pass CI until the updated ML.NET is available

isaacabraham:
Linking to #7

isaacabraham:
@dsyme haven't looked at this yet - what are the folder structures like in comparison to one another? I opted for an fsharp set of folders which had paket etc. in them - they were all self-contained. Has some pros and cons to that approach.

dsyme:
> @dsyme haven't looked at this yet - what are the folder structures like in comparison to one another? I opted for an fsharp set of folders which had paket etc. in them - they were all self-contained. Has some pros and cons to that approach.

`csharp` and `fsharp` are parallel.  No use of `paket` here, the projects are compiled for now.

dsyme:
Closing in favour of #35  and #36

PR #35, opened 2018-08-03
move C# --> csharp

dsyme:
Move the C# samples to a `csharp` directory as discussed in #32

PR #36, opened 2018-08-03
Add F# samples

dsyme:
Variation on #32 to add F# samples after moving C# samples to "csharp" directory

Things to do:
* [ ] Depends on ML./NET containing https://github.com/dotnet/machinelearning/pull/600 and https://github.com/dotnet/machinelearning/pull/616
* [ ] Make sure the samples run (some "force modules to load" additions may be needed)
* [ ] Consider using XPlot as it is more normal for F#
* [ ] Consider making available as F# scripts rather than just compiled code (these would run on .NET Framework).  Adding a pivot "compiled" and "scripting" may make senses, I'm not sure.
* [ ] remove stuff that shouldn't be here like the change to nuget.config


CESARDELATORRE:
Would it be possible to add the F# samples (this PR) to the branch "dotnet/machinelearning-samples" instead of MASTER branch? - Until F# is supported in the public release of ML.NET.

I'll add the Csharp changes about the folders first, to MASTER and to "dotnet/machinelearning-samples" branches, though.

CESARDELATORRE:
Never mind, I just saw your email. Let's keep the PR against MASTER. But we'll need to wait for the merge to happen.

dsyme:
@ankitbko @CESARDELATORRE This change is now much smaller now #35 is in, if you'd like to start reading the F#. Main points to note are

* Each `Program.fs` is written in top-to-bottom script form, e.g. [like this](https://github.com/dotnet/machinelearning-samples/pull/36/files#diff-7d94ecda4d521ba8f3ab6293fbd8c03eR1).  

* We may eventually change them to be "script.fsx" files but for now they are compiled, for now that's fine

* When reading the code, you start reading at the top and go through to the bottom just like a python script.   Functions are defined before they get called.

* I've just used a single file for each example. F# people tend to put more in a file since definitions are generally more succinct.

* The data types are defined using `member val`.  F# people would normally like to use records without any attributes and it still feels a little odd to use classes and attributes in F#.  It would be nice to look at whether this can become simpler

* I think the plotting and CSV data access can be simplified, to use XPlot and FSharp.Data respectively, I'll look into that.

Good intro guides to F# are at https://fsharpforfunandprofit.com/

ankitbko:
@dsyme Pretty sure I am not the person you intended to tag 😄

dsyme:
@ankitbko  - You're right, was after @asthana86 :)

samples/nuget.config
```diff
@@ -5,5 +5,6 @@
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <!-- Use this if you want consume the daily builds of ML.NET -->
     <!-- <add key="Dailies" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" /> -->
+    <add key="localPath" value="C:\GitHub\dsyme\machinelearning\bin/packages" />
```
eerhardt: This should be reverted.  It is causing CI to fail.

PR #39, opened 2018-08-09
Fixed F# samples

### LINKED ISSUE #38: 3 F# samples fail due to unhandled exceptions
Hopefully I miss something obvious, please point me to the right direction.
3 F# samples fail due to Unhandled System.ArgumentOutOfRangeException.

Namely,

### BinaryClassification_SentimentAnalysis

It works fine.

### Clustering_Iris, MulticlassClassification_Iris

They fail:

    > dotnet run
    =============== Training model ===============
    Automatically adding a MinMax normalization transform, use 'norm=Warn' or 'norm=No' to turn this behavior off.

    Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentOutOfRangeException:
     Training feature column 'Features' must be a known-size vector of R4, but has type: Vec<R8, 4>.
    Parameter name: data

### Regression_TaxiFarePrediction

It fails:

    > dotnet run
    =============== Training model ===============

    Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentOutOfRangeException:
     Column 'Features' has invalid source types: All source columns must have the same type. Source types: 'Vec<R4, 3>, Vec<R4, 7>, R8, R8, Vec<R4, 6>'.
    Parameter name: Column

ncave:
Fixes #38

### DIFF
--- a/samples/fsharp/getting-started/BinaryClassification_SentimentAnalysis/Program.fs
+++ b/samples/fsharp/getting-started/BinaryClassification_SentimentAnalysis/Program.fs
@@ -14,16 +14,16 @@ type SentimentData() =
     member val SentimentText: string = "" with get, set
 
     [<Column("1", name="Label")>]
-    member val  Sentiment : double = 0.0 with get, set
+    member val Sentiment: float32 = 0.0f with get, set
 
 type SentimentPrediction() =
     [<ColumnName("PredictedLabel")>]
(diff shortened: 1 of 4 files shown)


dsyme:
These fixes look correct.  @CESARDELATORRE We should add CI testing that actually executes the samples?

PR #78, opened 2018-10-29
EshopDashboard Readme

taranais:
Update readme layout


samples/csharp/end-to-end-apps/eShopDashboardML/README.md
```diff
@@ -33,15 +33,138 @@ Learn how to set it up in Visual Studio plus further explanations on the code:
...
+
+ //model.Predict() predicts the nextperiod/month forecast to the one provided
+ ProductUnitPrediction prediction = predictor.Predict(dataSample);
+ Console.WriteLine($"Product: {dataSample.productId}, month: {dataSample.month + 1}, year: {dataSample.year} - Real value (units): 551, Forecast Prediction (units): {prediction.Score}");
+```
+
+
+#### 4. Consume the model
+
+Basically, we load the model, then the data file and finally we make a prediction function.
```
CESARDELATORRE: Here it is important to point out that the "consume the model" code is the code you actually use in the ASP.NET Core Web app. Then, also provide here what specific .cs file has this code within the ASP.NET Core web app.

Issue #83, opened 2018-10-29
F# samples need to be migrated to the new "Dynamic API available since v0.6"


CESARDELATORRE:
All the F# samples in this repo are still using the deprecated LearningPipeline API and need to be migrated to the new "Dynamic API", like we're doing with the C# samples.

Contributions from the community will be very much welcomed! 👍

mariuszwojcik:
I plan to work on porting **BinaryClassification_CreditCardFraudDetection** for F#.

kunjee17:
@mariuszwojcik if that is ok with you I ll take regression taxi fare, regression bike sharing and Deep Learning Image Classification (this one is new I ll port from C#) .

kunjee17:
@mariuszwojcik I guess there is a language constraint to implement things in C# way. I did file a issue to in F# language repo. Hopefully things will be solved. Linking it [here](https://github.com/fsharp/fslang-suggestions/issues/162) for reference.

cc/ @CESARDELATORRE @dsyme

kunjee17:
@mariuszwojcik you can start working if you have solved the issue. And I'll pull it from there. I haven't changed much other that few c# bits I forgot to change. I'll continue from that. Thanks

kunjee17:
@mariuszwojcik I guess you solved it. Thanks man. I am on holidays for Diwali. I ll start working again from next week. So, let me know whatever is pending so will start from there.

mariuszwojcik:
@kunjee17 Happy Diwali !!! Have a great holidays. Will keep you posted with my progress.

CESARDELATORRE:
@mariuszwojcik - Closing this issue as the migration of the F# samples is work in progress thanks to you guys! 👍

dsyme:
@CESARDELATORRE How far through the migration work are we, do we have a tracking issue for what remains to be done?  I know I promised to help and I've noticed work progressing, just wanted to know what's left to do, thanks

PR #89, opened 2018-10-30
#83 Moved MulticlassClassification_Iris F# sample to new dynamic API.

mariuszwojcik:
Continued with Pipeline module introduced with `BinaryClassification_SentimentAnalysis` sample, which is supposed to make some operations on pipeline more F# friendly. This allows to pipe operations for creating pipeline and training model (idea is similar to Seq/List/Array modules which allow to pipe chain methods):
```fsharp
    let model = 
        env
        |> Pipeline.concatEstimator "Features" [| "SepalLength"; "SepalWidth"; "PetalLength"; "PetalWidth" |]
        |> Pipeline.append (SdcaMultiClassTrainer(env, SdcaMultiClassTrainer.Arguments(), "Features", "Label"))
        |> Pipeline.fit trainingDataView
```

For now I copied file from `BinaryClassification_SentimentAnalysis` sample and added `concatEstimator` function, but I am thinking about moving it outside and share between projects.


CESARDELATORRE:
@dsyme - Hey Don, when possible can you review/approve this PR?
Thanks a lot! 👍

PR #85, opened 2018-10-29
Migrated BinaryClassification_SentimentAnalysis F# sample to Dynamic API

mariuszwojcik:
Migration done by following C# sample.
I have created simple `Pipeline` module to make some operations on pipeline more F# friendly. This allows to **pipe** operations for creating pipeline and training model:

```fsharp
    let model = 
        env
        |> Pipeline.textTransform "Text" "Features"
        |> Pipeline.append (LinearClassificationTrainer(env, LinearClassificationTrainer.Arguments(), "Features", "Label"))
        |> Pipeline.fit trainingDataView
```
I plan to *enhance* this module as I continue with migration of other projects.

PR #90, opened 2018-10-30
#83 Moved Clustering_Iris F# sample to new dynamic API. 

mariuszwojcik:
Continued with Pipeline module introduced with BinaryClassification_SentimentAnalysis sample, which is supposed to make some operations on pipeline more F# friendly. This allows to pipe operations for creating pipeline and training model (idea is similar to Seq/List/Array modules which allow to pipe chain methods):
```fsharp
    let model = 
        env
        |> Pipeline.concatEstimator "Features" [| "SepalLength"; "SepalWidth"; "PetalLength"; "PetalWidth" |]
        |> Pipeline.append (KMeansPlusPlusTrainer(env, "Features", clustersCount = 3))
        |> Pipeline.fit trainingDataView
```

For now I copied file from BinaryClassification_SentimentAnalysis sample and added concatEstimator function, but I am thinking about moving it outside and share between projects.

Applied code style suggestions from #85.

### DIFF
--- a/samples/fsharp/getting-started/Clustering_Iris/Clustering_Iris.fsproj
+++ b/samples/fsharp/getting-started/Clustering_Iris/Clustering_Iris.fsproj
@@ -9,6 +9,7 @@
     <Compile Include="Program.fs" />
     <Folder Include="datasets\" />
     <None Include="..\..\..\..\datasets\iris-full.txt" Link="datasets\iris-full.txt" />
+    <None Include="README.md" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


CESARDELATORRE:
@dsyme - Hey Don, when possible can you review/approve this **other** PR?
Thanks a lot! 👍

CESARDELATORRE:
I'm merging this PR so it is migrated to the dynamic API, but please, afterwards, when possible for you, review the C# code we have for Clustering_Iris as we are splitting the dataset so it trains just with 80% then tests and gets the metrics with the 20%, so we make the F# code similar.

mariuszwojcik:
@CESARDELATORRE thanks. In the next few days I will update the code as well as project / folder structure to what is in C# samples.

CESARDELATORRE:
@mariuszwojcik - Thanks a lot! You're a champion for the .NET team! 👍

PR #91, opened 2018-10-31
Added Project Structure Guidelines .md document

CESARDELATORRE:
Doc includes the following topics to discuss:
- Opinionated "Directory/folders structure" and reasons why we're recommending it
- Phases/sections for internal code to implement within a "model training/test" application

### DIFF
--- a/docs/STRUCTURE-GUIDELINES.md
+++ b/docs/STRUCTURE-GUIDELINES.md
@@ -227,3 +227,7 @@ Point to different samples on how to host trained models in an different applica
 #### Additional references for model builder app
 
 - [Machine Learning Project Structure: Stages, Roles, and Tools](https://www.altexsoft.com/blog/datascience/machine-learning-project-structure-stages-roles-and-tools/)
+
+## Opinionated code style practices
+
+TBD
--- a/docs/STRUCTURE-GUIDELINES.md
+++ b/docs/STRUCTURE-GUIDELINES.md
@@ -188,3 +188,42 @@ ML.NET Sample
 - [Organizing and testing projects with .NET Core](https://docs.microsoft.com/en-us/dotnet/core/tutorials/testing-with-cli)
 
 - [David Fowler's project structure (ASP.NET team)](https://gist.github.com/davidfowl/ed7564297c61fe9ab814)
+
+
+
+
+## Internal code structure for the Model Builder app
+ 
+
+### Phase A: Iterative process building the model
+
+	1. Input dataset (where is my data)
+	2. Loader function (define columns)
+	3. Feature engineering (process raw input data)
+	4. Learner (define my pipeline's model type & hyperparameters)
+	5. TrainTest / CV (get metrics for my pipeline)
+	6. (Look at metrics)
+	7. (Iterate to improve metrics -- goto step 3 or 4)
+	8. Re-train a model for production on 100% of data
+
+**Train with 100% of the data once the algorithm and config are decided**:
+ 
+You should do a full re-train after your iterative train/evaluate/test cycles because of the following reasons:
+
+- The full re-training on the entire dataset gives you a better model because it is trained with more data. 
+- The CV/TrainTest gives your pipeline's metrics { accuracy, AUC, NDCG, etc } so you can compare different approaches (algorithm/hyper-parameters). 
+- The full retraining, on 100% of the dataset, is the model to deploy in production.
+
+### B. Deploy the model to production application (end-user application)
+
+Point to different samples on how to host trained models in an different application types:
+- Web App (ASP.NET Core app, MVC, Razor)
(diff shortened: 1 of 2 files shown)


CESARDELATORRE:
Can you guys read it and provide feedback? We can have a meeting to discuss about it it needed.

docs/STRUCTURE-GUIDELINES.md
```diff
@@ -0,0 +1,233 @@
...
+├── docs                       <- Global docs folder documenting the whole sample
+├── images                     <- Images used by the .md documents
+│
+├── Model-Builder-App          <- Usually a console app for building/training/evaluating/testing a model
+│    ├── Data
+│    │   ├── Raw               <- The original, immutable data dump.
+│    │   ├── Working           <- Intermediate data that has been transformed.
+│    │   └── Processed         <- The final, canonical datasets for training/evaluating/testing.
+│    ├── Models                <- The trained, generated saved/serialized ML.NET model files
+│    ├── MyConsoleApp          <- Source code for this particular app
```
shmoradims: >MyConsoleApp [](start = 9, length = 12)

Is this extra level needed? It's not clear which of Model-Builder-App or MyConsoleApp is the project folder. (Project folders need to be called out)

docs/STRUCTURE-GUIDELINES.md
```diff
@@ -0,0 +1,233 @@
...
+│
+├── End-User-App               <- Actual final-user's app using the trained model just for predicting/scoring. 
+│    ├── MyWebApp              <- Source code for this particular app (WebApp, ConsoleApp, DesktopApp, etc.)
+│    │   ├── MLDataStructures  <- Data classes for observations and predictions (Remove if using shared library)
+│    │   ├── MLModels          <- ML.NET serialized .ZIP model files 
+│    │   ├── EntityModels      <- Entity classes when using databases, EF, etc. 
+│    │   └── Other files       <- App's files and folders
+│    └── MyTestProject         <- Unit test project 
+│
+└── Shared-Library             <- Shared Library with common classes across apps
```
shmoradims: >Shared-Library [](start = 4, length = 14)

This is ideally shared across all samples. For examples, a lot of the console helpers could be shared. We can create a separate projects, that other samples can add to their solution.

docs/STRUCTURE-GUIDELINES.md
CESARDELATORRE: I think we do. There are three concepts here. 
- 1. The logical end to end sample/application (e.g. CustomersSegmentation). I our case, this is a folder for the sample. If it were a real application, this would be a whole GitHub repo.
- 2. The specific logical App, either MLModelBuilder or the End-User-App. This includes assets like datasets, generated MLModels, or even databases, etc. out of the source-code folder.
- 3. The VS-code project , for instance:
  - CustomersSegmentation-console-app
  - Customers-Dasboard-web-app

docs/STRUCTURE-GUIDELINES.md
CESARDELATORRE: Mmmmm, the thing is that each sample should be autonomous. If you just copy any sample's folder to another machine, it should just work without additional dependencies to other projects.
 
This is kind of similar to a microservice autonomy goal. No shared code/libraries unless thy are truly generic like a NuGet package.

In the case of ConsoleHelpers, most of them are customized with specific "print" info related to each sample. So, I'm not sure about this.

If we identify a truly generic library/code (comparable to a JSON serializer NuGet package, meaning it is completely generic), we could do as you suggest and share it as a "Building Block". But only if needed. Dependencies across samples might be worse than good in many cases.

"Building Blocks" can dangerously evolve towards a "custom framework".. Not usually good for the enterprise, but a bottleneck.. ;) - But I'd be okay if we really need this concept.

CESARDELATORRE: We had a discussion about this. I agree on the point that for our own purposes, having a "shared-code" library with shared code across samples (e.g. ModelBuilder class, etc.) would help us on maintainability while migrating samples from one preview version to the next one, sine there are many breaking changes with each version... It would also provide consistency across samples, However, each sample is "a different app" and should be self-contained, autonomous. Having a "shared-code" library, kind of a "mini-framework", will make all the samples inter-related through a common point, which will make it rigid, not flexible, and if a user copies a sample folder, he will also need the additional "common-code" or "mini-framework" as a side-car code...

I'm going to do a proof of concept, since this approach suggested by Shahab can help us on the multiple migrations, but once we stabilize the ML.NET API, we might want to split it and really create fully autonomous samples for the sake of user's experience... "Custom frameworks" are not usually a good idea for independent applications, samples mimic independent applications...

Anyone else, do you have additional thoughts about this point? - The way we choose here will dictate a different way of implementing the samples moving forward...

PR #103, opened 2018-11-03
GitHub Issues Labeler sample updated to 0.7 and using "common-code" approach

CESARDELATORRE:
GitHub Issues Labeler sample updated to 0.7 and using "common-code" approach.
Also, related global refactoring for "common-code" approach.


samples/csharp/common/ConsoleHelper.cs
```diff
@@ -45,6 +45,30 @@ public static void PrintRegressionMetrics(string name, RegressionEvaluator.Resul
...
+            var logLoss          = metricsInMultipleFolds.Select(m => m.LogLoss);
+            var logLossReduction = metricsInMultipleFolds.Select(m => m.LogLossReduction);
+
+            Console.WriteLine($"**************************************************************************");
+            Console.WriteLine($"*       Metrics for {algorithmName} Multi-class Classification model      ");
+            Console.WriteLine($"*-------------------------------------------------------------------------");
+            Console.WriteLine($"*       Average MicroAccuracy:    {microAccuracies.Average():0.##}");
+            Console.WriteLine($"*       Average MacroAccuracy:    {macroAccuracies.Average():0.##}");
+            Console.WriteLine($"*       Average LogLoss:          {logLoss.Average():#.##}");
+            Console.WriteLine($"*       Average LogLossReduction: {logLossReduction.Average():#.##}");
```
justinormont: The command-line (maml) prints in this format:
```
Accuracy(micro-avg): 0.690777 (0.0106)
Accuracy(macro-avg): 0.447108 (0.0267)
Log-loss:           1.852149 (0.1039)
Log-loss reduction: 9.075382 (5.1809)

---------------------------------------
Physical memory usage(MB): 2825
Virtual memory usage(MB): 3564
6/4/2018 9:32:19 PM        Time elapsed(s): 138
```

The first value is the average, and second value in the parenthesis is the standard deviation of the CV folds. CV is a costlier mode vs TrainTest (n-times slower). The main gain in CV mode (vs TrainTest) is its better estimate of your metric and the error in the metric.

Can we print the standard deviation? 

---

Another interesting choice is a 95% [confidence interval](https://en.wikipedia.org/wiki/Confidence_interval#Basic_steps):
```
Accuracy(micro-avg): 0.690777 ± 0.0464
```

Basic formula: `Avg` ± `1.96 * StdDev / SQRT(numCVFolds)`. There are simplifying assumptions made, for instance assuming a normal distribution, but the utility is still there.

In the end, I use maml's printed standard deviation to calculate error bars, which lets me have a general idea if two runs are the same or one is statistically significantly better.

samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+            var dataProcessPipeline = dataProcessor.DataProcessPipeline;
+
+            // (OPTIONAL) Peek data (such as 2 records) in training DataView after applying the ProcessPipeline's transformations into "Features" 
+            Common.ConsoleHelper.PeekDataViewInConsole<GitHubIssue>(mlContext, trainingDataView, dataProcessPipeline, 2);
+            //Common.ConsoleHelper.PeekVectorColumnDataInConsole(mlContext, "Features", trainingDataView, dataProcessPipeline, 2);
+
+            // STEP 3: Set the training algorithm, then create and config the modelBuilder                 
+            var trainer = mlContext.MulticlassClassification.Trainers.StochasticDualCoordinateAscent("Label", "Features");
+            var modelBuilder = new Common.ModelBuilder<GitHubIssue, GitHubIssuePrediction>(mlContext, dataProcessPipeline); //, trainer);
+            modelBuilder.AddTrainer(new SdcaMultiClassTrainer(mlContext, "Features", "Label"));
```
justinormont: Recommend `OVA-AveragedPerceptron{iter=10}` for a text problem. It's our general winner across many text datasets, and therefore is recommended in example code.

samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/DataProcessor.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public IEstimator<ITransformer> DataProcessPipeline { get; private set; }
+
+        public DataProcessor(MLContext mlContext)
+        {
+            // Configure data transformations in the Process pipeline
+
+            DataProcessPipeline = new ValueToKeyMappingEstimator(mlContext, "Area", "Label")
+                                  .Append(new TextFeaturizingEstimator(mlContext, "Title", "TitleFeaturized"))
+                                  .Append(new TextFeaturizingEstimator(mlContext, "Description", "DescriptionFeaturized"))
```
justinormont: Recommend explicitly setting the ngram lengths to bigram+trichargrams:
https://github.com/dotnet/machinelearning/blob/7e5ee22c30d1184afa93a77a34de1aa7bb1e909b/test/Microsoft.ML.Tests/ScenariosWithDirectInstantiation/SentimentPredictionTests.cs#L51-L52

This both lets users know the parameters exist, and encourages users to modify them.

samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+            // in order to evaluate and get the model's accuracy metrics
+            Console.WriteLine("=============== Cross-validating to get model's accuracy metrics ===============");
+            var crossValResults = modelBuilder.CrossValidateAndEvaluateMulticlassClassificationModel(trainingDataView, 6, "Label");
+            ConsoleHelper.PrintMulticlassClassificationFoldsAverageMetrics("SdcaMultiClassTrainer", crossValResults);
+
+            // STEP 5: Train the model fitting to the DataSet
+            Console.WriteLine("=============== Training the model ===============");
+            modelBuilder.Train(trainingDataView);
+
+            // STEP 6: Save/persiste the trained model to a .ZIP file
```
justinormont: `persiste` should be `persist`

CESARDELATORRE:
Thanks a lot Justin for reviewing! - I'll take all of these points into account tomorrow. 👍

samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/Program.cs
CESARDELATORRE: 👍

samples/csharp/common/ConsoleHelper.cs
CESARDELATORRE: I added the standard deviation and 95% confidence interval for each metric.

This is how it'd be printed:
image

Canyou check if the following formulas in code right?

```
        public static double CalculateStandardDeviation (IEnumerable<double> values)
        {
            double average = values.Average();
            double sumOfSquaresOfDifferences = values.Select(val => (val - average) * (val - average)).Sum();
            double standardDeviation = Math.Sqrt(sumOfSquaresOfDifferences / values.Count());
            return standardDeviation;
        }

        public static double CalculateConfidenceInterval95(IEnumerable<double> values)
        {
            double confidenceInterval95 = 1.96 * CalculateStandardDeviation(values) / Math.Sqrt(values.Count());
            return confidenceInterval95;
        }

```

samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/DataProcessor.cs
CESARDELATORRE: For now, I think we prefer to keep the code simpler in most of the samples. But feel free to submit a PR about this point. We could add it as another possible more advanced and accurate configuration. 👍

samples/csharp/common/ConsoleHelper.cs
justinormont: Do we have either a standard deviation or preferably a standard-error (StdErr) function in our TLC stats package?  (or the [built-in](https://docs.microsoft.com/en-us/dotnet/api/system.data.objects.entityfunctions.standarddeviation?view=netframework-4.7.2))

If it's simple to call, it would be nice to re-use code. Someone likely has already thought of the corner cases. 

The confidence interval would be the full range of: `0.690777 ± 0.0464`. Also a leading `0` w/ four digits would look nice: `0.0150` (current is `.015`).

samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/Program.cs
justinormont: Correct, it needs to be used in an OVA (one-vs-all) to upgrade from binary to multi-class. The OVA is the form it was benchmarked in.

samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/DataProcessor.cs
justinormont: The added Bigram+Trichar isn't to have a better accuracy of the sample's dataset, simply the pedagogical route to have users make better models for their dataset. Users moving the n-gram lengths is often a quick route to gains, and the further hope of get users thinking about feature engineering (and not just learners + hyperparameters). 

You're right, it's a trade-off.

CESARDELATORRE: Feel free to add a small PR with this alternate code including n-gram lengths. At least, we could show it commented as another alternative for the dev/user to explore.

samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/Program.cs
CESARDELATORRE: My point is that I don't see an "OVA-AveragedPerceptron" trainer in the `mlContext.MulticlassClassification.Trainers` catalog, here:

image

Do you have any sample code for this approach using `OVA-AveragedPerceptron` for multi-class-classification?

Thanks! 👍

justinormont: Example of OVA-AveragedPerceptron: https://github.com/dotnet/machinelearning/blob/9d33efeed68e82ec67508117d239dfcdfa8121d9/test/Microsoft.ML.Tests/Scenarios/Api/CookbookSamples/CookbookSamples.cs#L620-L624

If you use AP for text, be sure to set iter=10 -- default is 1 iteration, which I've been meaning to revisit as a topic. We discussed it internally a year ago and, if I recall, we decided to move AP to 10 iterations though haven't acted on it.

PR #106, opened 2018-11-04
#83 Moved Regression_TaxiFarePrediction F# sample to dynamic API.

mariuszwojcik:

Refactored `Pipeline` module to separate project to allow reuse between projects (once all awaiting pull requests with sample migration are merged I will refactor them to use common project). The optimal solution would be to actually move that project to main ML.NET repository - @dsyme, what do you think about that approach?

Due to ML.NET code using generic variables which are not supported in F#, I have used *Type Test Pattern* matching in the `append` method.


CESARDELATORRE:
@dsyme - Hey Don, when possible can you review/approve this other PR on F#?
Thanks a lot! 👍

CESARDELATORRE:
@mariuszwojcik - When possible, can you compare it with the latest minor changes in TaxiFare in C# and create any new PR if still missing anything?

mariuszwojcik:
@CESARDELATORRE sure thing. I will update the project / folder to the format all samples are using now and bump it to use v0.7

Issue #110, opened 2018-11-05
Confidence Interval for CV metrics should be calculated using t-distribution quantiles


najeeb-kazmi:
In `samples\csharp\common\ConsoleHelpers.cs` , the 95% confidence interval upper bound for CV metrics is calculated using 1.96, which is the 97.5% quantile of the standard normal distribution. We should instead use the 97.5% quantile of the Student's t distribution with (n - 1) degrees of freedom, where n is the number of folds, because we use the sample standard deviation.

``` csharp
        public static double CalculateConfidenceInterval95(IEnumerable<double> values)
        {
            double confidenceInterval95 = 1.96 * CalculateStandardDeviation(values) / Math.Sqrt(values.Count());
            return confidenceInterval95;
        }
```

Moreover, the standard deviation calculation should use the formula for the sample instead of the population, i.e. use (n - 1) instead of n in the denominator.

``` csharp
        public static double CalculateStandardDeviation (IEnumerable<double> values)
        {
            double average = values.Average();
            double sumOfSquaresOfDifferences = values.Select(val => (val - average) * (val - average)).Sum();
            double standardDeviation = Math.Sqrt(sumOfSquaresOfDifferences / values.Count());
            return standardDeviation;
        }
```

@shmoradims @CESARDELATORRE @justinormont can we use either Accord.NET or Math.NET for the t-distribution inverse CDF function? Any potential licensing issues with either? Accord.NET is GNU Lesser GPL 2.1, Math.NET Numerics is MIT/X11.

CESARDELATORRE:
What are we using for this kind of formulas in MAML or TLC? It’d be great if we can use code from ML.NET.. Other than that, just for samples we can also use Math.NET..

PR #113, opened 2018-11-06
Fixing confidence Interval for CV metrics 

CESARDELATORRE:
Fixing confidence Interval for CV metrics should be calculated using t-distribution quantiles


CESARDELATORRE:
@najeeb-kazmi - Please, review this changes trying to fix the issues you mentioned here: https://github.com/dotnet/machinelearning-samples/issues/110

PR #146, opened 2018-11-29
Heart prediction example

MarcoZama:
This is a Multiclass example about Heart Disease. Dataset is this :https://archive.ics.uci.edu/ml/datasets/heart+Disease


CESARDELATORRE:
@MarcoZama - Thanks a lot for this sample. Before accepting the sample, could you update it so instead of the "Common.ModelBuilder" class it directly uses the ML.NET API like in the other samples? For instance, see the GitHub Multi-Class classification as a reference (you can use just SDCA, no need to use multiple optional algorithms).
 
Basically, we're stepping out of the "Common.ModelBuilder" class since we don't want folks to "learn" that wrapper class but the ML.NET API itself. 

QUESTION: What's the accuracy/metrics of your model?

Thanks! 👍

MarcoZama:
> @MarcoZama - Thanks a lot for this sample. Before accepting the sample, could you update it so instead of the "Common.ModelBuilder" class it directly uses the ML.NET API like in the other samples? For instance, see the GitHub Multi-Class classification as a reference (you can use just SDCA, no need to use multiple optional algorithms).
> 
> Basically, we're stepping out of the "Common.ModelBuilder" class since we don't want folks to "learn" that wrapper class but the ML.NET API itself.
> 
> QUESTION: What's the accuracy/metrics of your model?
> 
> Thanks! 👍

Thank you for the reply Cesar, 
i'll update and i'll send other Pull request!

CESARDELATORRE:
@MarcoZama - Will you be able to update this sample? Also, note that it should be updated to v0.9.
I see that it currently supports v0.7:

<PackageReference Include="Microsoft.ML" Version="0.7.0" />

Another possibility if you are not going to be able to update it and maintain it through PR is to publish a simple link at the [COMMUNITY-SAMPLES page here](https://github.com/dotnet/machinelearning-samples/blob/master/docs/COMMUNITY-SAMPLES.md), and point to any repo you might have? That way, it is less important if it is still not up-to-date to the latest ML.NET version.

Please, confirm.
Thanks! 👍

MarcoZama:
Yes! I’m doing now.

MarcoZama:
Close for [request 214](https://github.com/dotnet/machinelearning-samples/pull/214)!

PR #186, opened 2018-12-21
Added expirationTime and improvements to Object Pooling for PredictionEngine

CESARDELATORRE:
Added expirationTime and better management for minPoolSize and maxPoolSize

### DIFF
--- a/samples/csharp/common/ObjectPool.cs
+++ b/samples/csharp/common/ObjectPool.cs
@@ -8,21 +8,35 @@ namespace Common
 {
     public class ObjectPool<T>
     {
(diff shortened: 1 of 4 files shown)


samples/csharp/common/ObjectPool.cs
```diff
@@ -8,29 +8,44 @@ namespace Common
 {
     public class ObjectPool<T>
```
eerhardt: @CESARDELATORRE - why are we creating our own ObjectPool class here? Why not use https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.objectpool.objectpool-1 instead?

Issue #202, opened 2019-01-15
F# samples need to be migrated to ML.NET v0.9


mariuszwojcik:
Migrate F# samples to ML.NET 0.9.

kevmal:
There are no more F# samples pre version 0.9

CESARDELATORRE:
Closing this issue.
Thanks! :)

PR #203, opened 2019-01-15
#202: migrate BikeSharingDemand F# sample to v0.9

mariuszwojcik:
Simple migration of BikeSharingDemand F# sample to v0.9.
Common modules have been migrated to v0.9 in separate folder to keep projects not migrated yet working.


CESARDELATORRE:
@mariuszwojcik - The BikeSharingDemand sample has an issue (it also happens in C#).

Basically, in 0.9, pipeline cache is not enabled by default (because of multiple reasons) and there's a related bug when you run the sample in VS with F5 (Debugging attached to process) where too many threads are created and the process never-ends...

The bug is explained in this issue I opened in the ML.NET repo: https://github.com/dotnet/machinelearning/issues/2099

Basically, the workaround is to explicitly enable the pipeline cache with this:

`.AppendCacheCheckpoint(mlContext);`

You might notice a lot of training perf improvement when using the cache, too, so you might want to add it to additional samples if it improves the training and test time. :)

However, this change has to be done directly into the pipeline. You don't have that method in the "ModelBuilder" helper class.

I would recommend to step-out of the "ModelBuilder" helper class as we did in late November 2018 for the C# samples because these kind of code changes but mostly because we want developers to directly learn the ML.NET API. If using these helper classes, a developer might not clearly see how the ML.NET  API is shaped.. This is specially important for the pipeline-chain API.

Please, let's fix this issue when possible for you. Not in a rush though, since the sample works with out F5 but with Ctrl+F5. 👍 

Thank you! :)

Issue #213, opened 2019-01-24
F# BikeSharing sample never ends training when running as F5 in Visual Studio


CESARDELATORRE:
The F# BikeSharingDemand sample has an issue (it also happens in C# for other samples if not using a workaround explained below).

Basically, in 0.9, the pipeline cache is not enabled by default (because of multiple reasons) and there's a related bug when you run the sample in VS with F5 (Debugging attached to process) where too many threads are created and the process never-ends...

The bug is explained in this issue I opened in the ML.NET repo: 
https://github.com/dotnet/machinelearning/issues/2099

Basically, the workaround is to explicitly enable the pipeline cache with this:

.AppendCacheCheckpoint(mlContext);

You might notice a lot of training perf improvement when using the cache, too, so you might want to add it to additional samples if it improves the training and test time. :)

However, this change has to be done directly into the pipeline. You don't have that method in the "ModelBuilder" helper class.

I would also recommend to step-out of the "ModelBuilder" helper class as we did in late November 2018 for the C# samples because these kind of code changes but mostly because we want developers to directly learn the ML.NET API. If using these helper classes, a developer might not clearly see how the ML.NET API is shaped.. This is specially important for the pipeline-chain API.

mariuszwojcik:
@CESARDELATORRE thanks for info, I will look into this in the next few days.

And regarding 'ModelBuilder' helper module - makes sense! Will move the code into relevant functions.

kevmal:
I think this issues has been resolved

prathyusha12345:
Closing this issue. 
Thanks :)

@CESARDELATORRE Adding for referrence

PR #228, opened 2019-02-04
#213:  append CacheCheckpoint in F# samples to mitigate ML.NET bug #2099

mariuszwojcik:
@CESARDELATORRE I have added call to `.AppendCacheCheckpoint(mlContext);` to mitigate the problem. 
The problem doesn't occur on my PC. I am using VS 2017 v15.9.6 Community edition tough. Can you please check on your PC whether it works OK?


CESARDELATORRE:
The issue happens only in some samples/algorithms.
For instance, in C# it happens in the GitHubLabeler (multi-class) but also in the BikeSharing (regression)..
And sometimes it might not happen. IT is not 100% predictable.

PR #231, opened 2019-02-04
#FS230: move F# BikeSharingDemand example to use direct ML.NET API.

mariuszwojcik:


mariuszwojcik:
@CESARDELATORRE - done.

PR #129, opened 2018-11-14
WIP: Clustering Customer Segmantation 

kunjee17:

PR #236, opened 2019-02-06
customer segmentation sample ported to F#

kevmal:
I literally just noticed #129 is F# as well. But anyway... maybe it's still of use.


CESARDELATORRE:
@kevmal - Hey! Thanks for your PRs migrating to F#. About this PR, could you resolve the conflicts it has?
Thanks a lot! 👍

PR #240, opened 2019-02-07
spam detection sample ported to F#

kevmal:
Unlike the C# sample this does not use a custom mapping partly due to https://github.com/dotnet/machinelearning/issues/2422.


CESARDELATORRE:
Thank you for your contribution!! 👍

Issue #254, opened 2019-02-15
Update ALL samples to .NET Core 2.2


CESARDELATORRE:
Some samples are in .NET Core 2.0, 2.1, 2.2.
We need to make sure all of them are on .NET Core 2.2

CESARDELATORRE:
Updated to .NET Core 2.1 in this PR:
https://github.com/dotnet/machinelearning-samples/pull/255

We choose 2.1 because it's long term support versus 2.2 which is shorter support.

PR #255, opened 2019-02-15
Update samples to .NET Core 2.2 (from 2.0 and 2.1)

stephentoub:
cc: @CESARDELATORRE, @eerhardt

### DIFF
--- a/samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/GitHubLabeler.csproj
+++ b/samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/GitHubLabeler.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
     <SignAssembly>false</SignAssembly>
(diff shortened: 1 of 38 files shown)


eerhardt:
why not stick with `2.1`?  It is LTS, and will supported for longer than `2.2`.

stephentoub:
> why not stick with 2.1? It is LTS, and will supported for longer than 2.2.

My main goal was to get off 2.0, which isn't supported.  I chose 2.2 because it's the latest that's released.  But sure, we could stick with 2.1.

CESARDELATORRE:
Related to this issue: https://github.com/dotnet/machinelearning-samples/issues/254

stephentoub:
> We need to make sure all of them are on .NET Core 2.2

@CESARDELATORRE, per @eerhardt's feedback I updated my PR to move them all to 2.1 rather than 2.2.  Should they go to 2.2?

CESARDELATORRE:
It's okay to have them in 2.1. They are just console apps.
What we'll migrate to 2.2 are the ASP.NET Core apps because ASP.NET Core 2.2 has new important things.

CESARDELATORRE:
This PR broke the project due to change of .NET Core version. It should be .NET core 2.2 for this ASP.NET Core 2.2 sample end-to-end-apps\Recommendation-MovieRecommender\Recommendation_MovieRecommenderApp.sln  which is an ASP.NET Core 2.2 application so it cannot be downgraded to .NET Core 2.1
I fixed my returning to .NET Core 2.2. 👍

stephentoub:
If some projects in the repo need to be 2.2, seems like they all should be for consistency?

PR #257, opened 2019-02-15
Turn off supportSparse in GitHubLabeler sample

stephentoub:
The input isn't sparse, but the ReadFromTextFile is using the default of
supportSparse=true, which is causing the loader to try and fail to parse
every row as if it were sparse (you can see this if you hook up an
mlcontext.Log handler and print out every message containing "Format").

Related to https://github.com/dotnet/machinelearning/issues/2576
cc: @CESARDELATORRE, @eerhardt


CESARDELATORRE:
Questions:
- Is this improving the performance (training time)?
- If this is the by default if not provided and we now need to provide "supportSparse: false", is that the most common case? - My point is, should this parameter be false for most of the samples? if yes, why is supportSparse not defaulted to false in the API?

I find this parameter to be "yet another" concept in the "by default" API...

Issue #264, opened 2019-02-21
F# TensorFlow Estimator sample shows Exception in the console output


prathyusha12345:
@kevmal The F# sample TensorFlow Estimator shows exception in the console. Please see the below output when running the F# sample

![image](https://user-images.githubusercontent.com/22335043/53189278-fe1d4b80-35bb-11e9-9f57-f3158c8e07cf.png)

Expected result(what you get with C# sample):
---------------------------------------
![image](https://user-images.githubusercontent.com/22335043/53189829-2fe2e200-35bd-11e9-9e81-f7189227a66c.png)

prathyusha12345:
Fixed in this PR https://github.com/dotnet/machinelearning-samples/pull/272

prathyusha12345:
Reopening as this issue is not fixed yet. Attached another PR wrongly.

kevmal:
@prathyusha12345  Are you still getting an exception after #271?

prathyusha12345:
Now I don't see any exception now. Thanks for fixing it. I am closing the issue.

PR #271, opened 2019-02-22
create output directory for F# tensorflow estimator sample

kevmal:
output directory needs to exist for tensorflow estimator sample. #264

### DIFF
--- a/samples/fsharp/getting-started/DeepLearning_TensorFlowEstimator/ImageClassification.Train/assets/outputs/README.txt
+++ b/samples/fsharp/getting-started/DeepLearning_TensorFlowEstimator/ImageClassification.Train/assets/outputs/README.txt
@@ -1,0 +1,1 @@
+output gets saved to this directory


samples/fsharp/getting-started/DeepLearning_TensorFlowEstimator/ImageClassification.Train/assets/outputs/README.txt
```diff
@@ -0,0 +1 @@
+output gets saved to this directory
```
prathyusha12345: @CESARDELATORRE  we don't have this file in C# sample. I think the name **outputs** itself describes the output files. I feel no need to add this file. Cesar Please add your inputs.

CESARDELATORRE: I put it so the folder gets created even if there's no model .ZIP file in there.. But I agree that this might not be needed. I'll merge it in any case. Thanks! 👍

kevmal: Was only to force the directory to be created in the repo, following the example of other projects (for example outputs directory of customer segmentation). Though the usual message is typically
```
This folder is empty until you run the training console app and create/train and save the models as .ZIP files.
Those model .ZIP files should appear in this folder.
```
Alternatively, could add a .gitignore to force the directory or add a `CreateDirectory` to the code.

PR #277, opened 2019-02-25
Refactor path.combine to remove + symbol

prathyusha12345:
Removed + symbol in path.Combine of in asll samples as it is not required.
Replaced ..\ with ../ to be compatible with Linux/MAc O.S
Removed Common folder in MovieRecommendation sample which I have added previously while fixing Path.

### DIFF
--- a/samples/csharp/getting-started/MulticlassClassification_Iris/IrisClassification/IrisClassificationConsoleApp/Program.cs
+++ b/samples/csharp/getting-started/MulticlassClassification_Iris/IrisClassification/IrisClassificationConsoleApp/Program.cs
@@ -12,12 +12,17 @@ namespace MulticlassClassification_Iris
     {
         private static string AppPath => Path.GetDirectoryName(Environment.GetCommandLineArgs()[0]);
 
-        private static string BaseDatasetsLocation = @"../Data";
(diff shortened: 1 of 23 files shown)


samples/csharp/getting-started/MulticlassClassification_Iris/IrisClassification/IrisClassificationConsoleApp/Program.cs
```diff
@@ -48,10 +48,10 @@ private static void BuildTrainEvaluateAndSaveModel(MLContext mlContext)
             
 
             // STEP 2: Common data process configuration with pipeline data transformations
-            var dataProcessPipeline = mlContext.Transforms.Concatenate(DefaultColumnNames.Features, "SepalLength",
-                                                                                   "SepalWidth",
-                                                                                   "PetalLength",
-                                                                                   "PetalWidth")
+            var dataProcessPipeline = mlContext.Transforms.Concatenate(DefaultColumnNames.Features, nameof(IrisData.SepalLength),
```
prathyusha12345: Refactored so that we can avoid runtime errors by replccing strings with strong types. I missed it in v0.10 migraion.

PR #288, opened 2019-03-01
F# - WorkingDay was missing from feature set for Bike sample

kevmal:
Also a couple of very small console output changes to match C# samples.


prathyusha12345:
Tested the PR.Working fine. Thanks @kevin for your changes. I am merging the PR.

PR #290, opened 2019-03-01
Remove duplicate icons for C#/F# samples

kevmal:
Just noticed my last update was duplicating icons. This just removes the extra getting-started/end-to-end icon to be consistent with the other links.


prathyusha12345:
@kevmal  thanks for the changes. I am merging the PR.

PR #295, opened 2019-03-04
Database Example

### LINKED ISSUE #143: Database example
### Purpose
There are requests for ML.Net to support  database integration to retrieve data. While we do not have database support, we do provide a way to create an IDataView from an IEnumerable, which provides an open-ended way to integrate data into an ML.Net pipeline. 

Therefore I am proposing that we provide an example of how to integrate a database as a datasource into an ML.Net pipeline using ML.Net's interface to create an IDataView from an IEnumerable. A working example would give users a starting point for their own integration. Having an example will allow us to understand the performance and limitations with using a database as a datasource.

### Entity Framework
Rather than provide a database specific implementation (like ODBC or SQL), this example will use the Entity Framework. The Entity Framework would cover a number of databases and is cross plat as it is supported in .net core. This also provides a common framework that .net developers will most likely be familiar with.

Entity Framework will also give configuration options that we can experiment with to determine what provides the best performance when using ML.Net. An example of this is eager loading vs lazy loading. 

More information about the Entity Framework can be found here:
https://docs.microsoft.com/en-us/ef/

### References
Here are the issues from ML.Net requesting/asking about database support:
https://github.com/dotnet/machinelearning/issues/1130
https://github.com/dotnet/machinelearning/issues/107
https://github.com/dotnet/machinelearning/issues/96

singlis:
The sample provides a solution that integrates with the existing sample
solution, along with instructions for how to create the database and run
the example.

This fixes #143


samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/DatabaseIntegration.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="EntityFramework" Version="6.2.0" />
```
divega: This reference shouldn't be here. I am surprised that nothing complains given that EF6.2 isn't compatible with .NET Core.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Model.cs
```diff
@@ -0,0 +1,57 @@
...
+        //public DbSet<Blog> Blogs { get; set; }
+        //public DbSet<Post> Posts { get; set; }
+        public DbSet<AdultCensus> AdultCensus { get; set; }
+
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            optionsBuilder.UseSqlite("Data Source=mlexample.db");
+        }
+    }
+    public class AdultCensus
```
divega: NIT: if you rename this class to Adult, AdultId will be recognized as a database generated key by convention (because it matches the naming pattern classname + "Id") and the attributes won't be necessary.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Model.cs
```diff
@@ -0,0 +1,57 @@
...
+        public string Sex { get; set; }
+        public string CapitalGain { get; set; }
+        public string CapitalLoss { get; set; }
+        public int HoursPerWeek { get; set; }
+        public string NativeCountry { get; set; }
+        public int Label { get; set; }
+    }
+
+
+    public class Blog
```
divega: Are Blog and Post just leftovers?

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        public static string DownloadAdultDataset()
+            => Download("https://raw.githubusercontent.com/dotnet/machinelearning/244a8c2ac832657af282aa312d568211698790aa/test/data/adult.train", "adult.txt");
+
+        public static List<AdultCensus> GetAdultData()
+        {
+            string adultDataset = DownloadAdultDataset();
+            string dataLine;
+            var dataset = new List<AdultCensus>();
```
divega: I believe you could easily refactor this to add the AdultCensus objects directly to the DbContext, eliminating the need for `List<AdultCensus>`, and the code would even look a bit simpler.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+                        NativeCountry = data[13],
+                        Label = int.Parse(data[14])
+                    });
+                }
+            }
+
+            return dataset;
+        }
+
+        public static void PopulateDatabase()
```
divega: NIT: You could use the word "seeding" instead of populating the database. It is a term familiar to many EF Core users.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            var adultData = GetAdultData();
+            using (var db = new AdultCensusContext())
+            {
+                foreach(var row in adultData)
```
divega: I still prefer the option of adding things directly to the DbContext and skipping the list, but if you have a list, or an IEnumerable (e.g. a LINQ query that reads the stream and projects AdultCensus instances with a Select(), you can replace this loop with simply db.AdultCensus.AddRange(adultData).

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+                }
+            }
+
+            return dataset;
+        }
+
+        public static void PopulateDatabase()
+        {
+            var adultData = GetAdultData();
+            using (var db = new AdultCensusContext())
```
divega: I believe somewhere you should be calling db.Database.EnsureCreated(). Opening the connection can create the database file automatically, but I believe you need this to make sure the tables are created as well.

divega: You may choose to call EnsureDeleted() before that as well, if you want the sample to re-download the database every time it runs.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+
+        public static void Main()
+        {
+            PopulateDatabase();
+            using(var db = new AdultCensusContext())
+            {
+                var mlContext =  new MLContext();
+                var dataView = mlContext.Data.ReadFromEnumerable(db.AdultCensus);
```
divega: use db.AdultCensus.AsNoTracking() here to get a stream of objects. Otherwise the DbContext will unnecessarily start tracking every object returned by the query, and all the data will need to fit in memory at the same time. 

I get that this is a bit artificial in this case because we hold all the data in a `List<T>`, but I think it is still important to recommend [no tracking queries](https://docs.microsoft.com/en-us/ef/core/querying/tracking) in case customers want to read larger datasets using EF Core.

divega: I wonder if you could do something a bit more interesting here, e.g. add a .Where() operator that allows selecting only some subset of the data from the command line. Or random sorting on the database (assuming random sorting is useful in this case).

samples/csharp/getting-started/DatabaseIntegration/README.md
```diff
@@ -0,0 +1,28 @@
...
+This sample demonstrates how to use a database as a data source for an ML.Net pipeline using the IDataView interface. The data can then be used for training and testing a model. In order
+to support mulitple databass the Entity Framework is used as part of this sample.
+
+## Problem
+
+In scenarios where the dataset lives in a database, the dataset is likely large and associated but not live within a single table. In addition, having to massage the data to be consumeable into an machine learning pipeline can be an expensive and tedious chore.
+
+
+## Solution
+This sample shows how to use the Entity Framework to connect to a database and leverage the existing schema allowing to query for the data that is needed to train and make this available to ML.Net through the IDataView interface.
```
divega: Say "Entity Framework Core" here, otherwise it is ambiguous.

samples/csharp/getting-started/DatabaseIntegration/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This sample uses SQLite to help demonstrate the database integration, but any database that is supported by the Entity Framwork can be used. This sample also uses ML.Net's ability to convert an IEnumerable to and IDataView -- as the Entity Framework will expose the results of a query as an IEnumberable. To get started the following commands need to run to setup the database. To create the database, the dot net migrations commands are used.
```
divega: I would then mention that we used a no tracking query to not require all data from being loaded in memory at the same time here.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+        public static List<AdultCensus> GetAdultData()
+        {
+            string adultDataset = DownloadAdultDataset();
+            string dataLine;
+            var dataset = new List<AdultCensus>();
+            using (StreamReader reader = new StreamReader(adultDataset))
+            {
+                // read the header
+                dataLine = reader.ReadLine();
+                while ((dataLine = reader.ReadLine()) != null)
```
divega: I prefer how @jwood803 accomplished this using LINQ in https://github.com/dotnet/machinelearning/pull/2902/files#diff-8696be27c85dee937c625ba6c27d564fR55, but I would start with ReadLines instead of ReadAllLines.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
divega: This makes me wonder if there would be value in having a method in the example that encapsulates how to get an IDataView from and EF Core query, e.g. pushing the AsNoTracking() and the optional ability to do some randomization there.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Model.cs
singlis: Yes they are, I will remove

---
In reply to: [270254655](https://github.com/dotnet/machinelearning-samples/pull/295#discussion_r270254655) [](ancestors = 270254655)

samples/csharp/getting-started/DatabaseIntegration/README.md
singlis: Diego - I found this URL that talks about NoTracking, but it discusses about how its great for read-only scenarios in that it does not have to track changes on each item. However, I don't see any mention about it not requiring all data being loaded into memory at the same time. Do you have more references to this that I can refer to in the example?
[https://docs.microsoft.com/en-us/ef/core/querying/tracking](https://docs.microsoft.com/en-us/ef/core/querying/tracking)


---
In reply to: [270257293](https://github.com/dotnet/machinelearning-samples/pull/295#discussion_r270257293) [](ancestors = 270257293)

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
singlis: Take a look at what I have now and see if this is acceptable as this is much more simplified. So I was not able to put the insert into the Select as the insert (or AddRange) is needed for the whole thing to become "realized". So created the data variable is just an enumerable and wont be evaluated until addrange is called.

---
In reply to: [270255775](https://github.com/dotnet/machinelearning-samples/pull/295#discussion_r270255775) [](ancestors = 270255775)

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
singlis: Also left feedback in another comment about NoTracking.

---
In reply to: [270256412](https://github.com/dotnet/machinelearning-samples/pull/295#discussion_r270256412) [](ancestors = 270256412)

singlis: So what I have now is that we query all the data from the table and use that for training. Then query again only selecting the top 2000 rows to use for prediction. Does this seem reasonable? Or are you wanting something more precise in that we always select one portion of the data for training and another portion for testing?

---
In reply to: [270259895](https://github.com/dotnet/machinelearning-samples/pull/295#discussion_r270259895) [](ancestors = 270259895)

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/DatabaseIntegration.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
divega: I think if we go with EF Core 3.0 previews (not a bad idea), we should switch to netcoreapp3.0 as well. Right now things may work, but in an upcoming preview this combination will stop working.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/DatabaseIntegration.csproj
```diff
@@ -0,0 +1,22 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="3.0.0-preview.19074.3" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="3.0.0-preview.19074.3" />
+    <PackageReference Include="Microsoft.ML" Version="$(MicrosoftMLVersion)" />
```
divega: How is MicrosoftMLVersion set? Does this prevent anyone from just cloning and running the project as normal? I suspect either having fixed dependencies and updating the projects to use the latest, or using '1.0.0*'s would probably be better.

samples/csharp/getting-started/DatabaseIntegration/README.md
divega: I was purposedly providing you with a bit more information than what we currently have in the documentation :smile: 

The truth is that no tracking queries in EF Core 1.x and 2.x are not always streaming and the rules are tricky. That is part of what I wrote in https://github.com/dotnet/machinelearning-samples/issues/143#issuecomment-443410619. In EF Core 3.0 this is getting easier to predict, because we reverted some of the previous decisions:

1. We will generate only one SQL query for each LINQ query, regardless of whether you have Include()
2. We don't do any identity resolution, which means that we don't keep weak references to the entities already returned.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <param name="dataFile">The location of where to save the downloaded file.</param>
+        /// <returns></returns>
+        private static string Download(string uriPath, string dataFile)
+        {
+            using (WebClient client = new WebClient())
+            {
+                client.DownloadFile(new Uri($"{uriPath}"), dataFile);
+            }
+
+            return dataFile;
```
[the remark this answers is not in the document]
divega: I think you could also read the data from the URL into the DbContext directly without going through a file. Searching I fould code like this (but I haven't tested it):
``` csharp
public IEnumerable<string> ReadLines(string url)
{
  using(var client = new WebClient())
  using(var stream = client.OpenRead(url))
  using(var reader = new StreamReader(stream))
  {
    string line;
    while ((line = reader.ReadLine()) != null)
    {
      yield return line;
    }
  }
}
```

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+                // our scenarios our read-only with the data. Not having tracking will  
+                db.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
+                var mlContext =  new MLContext();
+               
+                // Query our training data from the database. This query is selecting everything from the AdultCensus table. The
+                // result is then loaded by ML.Net through the LoadFromEnumerable. LoadFromEnumerable returns an IDataView which
+                // can be consumed by an ML.Net pipeline.
+                // NOTE: For training, ML.Net requires that the training data is processed in the same order to produce consistent results.
+                // Therefore we are sorting the data by the AdultCensusId, which is an auto-generated id.
+                var dataView = mlContext.Data.LoadFromEnumerable(db.AdultCensus.OrderBy(x=>x.AdultCensusId).ToList<AdultCensus>());
```
divega: `ToList<AdultCensus>()` will bring all the data into memory, defeating everything else we are doign to avoid that. Just pass the query (which is already an `IEnumerable<AdultCensus>()`) to the next API.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                Console.WriteLine("Training model...");
+                var model = pipeline.Fit(dataView);
+
+                Console.WriteLine("Predicting...");
+                // Now that the model is trained, we want to test it's prediction results, which is done by using a test dataset
+                // and transforming it by the trained model.
+                // To retrieve our test data, this issues a second query, retrieving the first 2000 records, then converts to an IDataView
+                // in order to feed into the ML.Net pipeline.
+                var testDataView = mlContext.Data.LoadFromEnumerable(db.AdultCensus.Take(2000).ToList<AdultCensus>());
```
divega: Idem with the ToList here, althought it is less important because you have already limited the size of the sample with Take.

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/DatabaseIntegration.csproj
singlis: Does going with .netcoreapp3.0 force users to use VS2019? I would prefer to have an update like that happen as a change across all projects in the sample. I will add the EFcore 3.0 reference though. 

---
In reply to: [278283185](https://github.com/dotnet/machinelearning-samples/pull/295#discussion_r278283185) [](ancestors = 278283185)

samples/csharp/getting-started/DatabaseIntegration/README.md
```diff
@@ -0,0 +1,33 @@
...
+This sample shows how to use the Entity Framework Core to connect to a database, query  and feed the resulting data into an ML.Net pipeline.
+
+This sample uses SQLite to help demonstrate the database integration, but any database that is supported by the Entity Framwork Core can be used. As ML.Net can  create an IDataView from an IEnumerable, this sample will use the IEnumerable that is returned from a query to feed the data into the ML.Net pipeline. To prevent the Entity Framework Core from loading all the data in from a result, a no tracking query is used. 
+
+To get started the following dot net migration commands need to run to setup the database. 
+
+### 1. Create the database
+From the command line in the directory where the DatabaseIntegration.csproj file is located, type the following commands:
+```
+dotnet ef migrations add InitialCreate
```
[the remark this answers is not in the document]
singlis: This is awesome, I am glad that users dont have to run these commands.

---
In reply to: [278292463](https://github.com/dotnet/machinelearning-samples/pull/295#discussion_r278292463) [](ancestors = 278292463)

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/DatabaseIntegration.csproj
singlis: This is a great question. I was following what other projects used,  thinking this was defined in a central place (i.e. I was hoping it was a props file that we include), however when looking for where this is defined, I found this:
[https://github.com/dotnet/machinelearning-samples/blob/e78e227d0269accbc74db26a1edffd0b782d2511/samples/csharp/getting-started/DeepLearning_ObjectDetection_Onnx/Directory.Build.props#L4](https://github.com/dotnet/machinelearning-samples/blob/e78e227d0269accbc74db26a1edffd0b782d2511/samples/csharp/getting-started/DeepLearning_ObjectDetection_Onnx/Directory.Build.props#L4)

I dont understand how this is exactly working. I am setting to 1.0.0-preview for now and will likely file an issue on this. 



---
In reply to: [278285209](https://github.com/dotnet/machinelearning-samples/pull/295#discussion_r278285209) [](ancestors = 278285209)

samples/csharp/getting-started/DatabaseIntegration/DatabaseIntegration/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+        public static void Main()
+        {
+            // Seed the database with the dataset.
+            CreateDatabase(datasetUrl);
+
+            using(var db = new AdultCensusContext())
+            {
+                // Set the query tracking behavior to be NoTracking, this is particularly useful because
+                // our scenarios only require read-only access.
+                db.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
```
divega: I have a slight preference for having AsNotracking() call in-line in the query instead. My reasoning:

- It is a simpler API 
- Presumably some customers will want to use some data in their database to train an ML model, while keeping the rest of their EF Core usage untouched
- If you are copying and pasting the key code from this sample into your own app, the chances that you will copy AsNoTracking() seem a bit higher

prathyusha12345:
@singlis can you take latest of master and push the changes to resolve conflicts.

singlis:
Closing as this PR was copied and merged. Please see #405

PR #296, opened 2019-03-05
Revert "Turn off supportSparse in GitHubLabeler sample"

stephentoub:
Reverts dotnet/machinelearning-samples#257

This is no longer needed, as:
- https://github.com/dotnet/machinelearning/pull/2630 made supportSparse default to false.  And even if it wasn't changed....
- https://github.com/dotnet/machinelearning/pull/2579 made supportSparse faster when it is true, in particular when the debugger is attached.

cc: @CESARDELATORRE

### DIFF
--- a/samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/Program.cs
+++ b/samples/csharp/end-to-end-apps/MulticlassClassification-GitHubLabeler/GitHubLabeler/GitHubLabelerConsoleApp/Program.cs
@@ -57,7 +57,7 @@ namespace GitHubLabeler
             var mlContext = new MLContext(seed: 1);
 
             // STEP 1: Common data loading configuration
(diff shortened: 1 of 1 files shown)


CESARDELATORRE:
> Reverts #257
> This is no longer needed, as:
> 
> dotnet/machinelearning#2630 made supportSparse default to false.  And even if it wasn't changed....
> dotnet/machinelearning#2579 made supportSparse faster when it is true, in particular when the debugger is attached.
> 
> cc: @CESARDELATORRE

But that is for v0.11, right? - We have the v0.11 migration (work in progress) in this other branch:
https://github.com/dotnet/machinelearning-samples/tree/migration/v.11 

Master branch is still v0.10 until v0.11 NuGet package is released at the main NuGet feed.

stephentoub:
> But that is for v0.11, right?

Yes.

> We have the v0.11 migration (work in progress) in this other branch:

Ok.  Would you like to:
a) Leave this open and merge it once 0.11 is released?
b) Merge it now anyway? (It really only affects perf while the debugger is attached, and this simplifies the sample.)
c) Close it and have me open a new PR to that other branch?

CESARDELATORRE:
Agree. This is not really impacting strongly on 0.10. I'll merge here and it'll be better by updating to 0.11 when we merge 0.11 branch. Thanks! 👍

PR #299, opened 2019-03-06
F# - Migrate MachineLearning-Samples to v0.11

kevmal:
tensorflow samples are set to version 0.10


prathyusha12345:
@kevmal When I ran F# CreditCardFraudDetection sample, it shows the predicted label as **false** for 2 transactions and score in negative values with labels **true**. I compared some of rows in  test.csv file in c# with f#  samples.I found little difference. so I copied the c# test.csv into f# samples folder. But still i see the same result of showing **false** predictedlabel. 

Can you please check once.

F# sample output:
-------------------
image

c# sample output:
-------------------
image

@CESARDELATORRE adding Cesar for reference.

prathyusha12345:
@kevmal F# TensorFlowEstimator-->ImageClassification.Train is missing predictedlabel in output messages.  For example **ImagePath: broccoli.jpg predicted as  with score 0.952327**. It should be **ImagePath: broccoli.jpg predicted as _broccoli_ with score 0.952327**

image

kevmal:
Updated the following
- TF sample needed to explicitly specify "PredictedLabelValue" as the output
- Credit card fraud C# sample was training on the test set which was causing the prefect results

kevmal:
Any other comments on this?

prathyusha12345:
@kevmal Got busy with other tasks. let me check now.

prathyusha12345:
@kevmal I still see the output for **CreditCardFraudDetection** is showing output incorrectly. i.e for Label=true, PredictedLabel = false. can you check once

image

samples/csharp/getting-started/BinaryClassification_CreditCardFraudDetection/CreditCardFraudDetection.Trainer/Program.cs
```diff
@@ -21,8 +21,8 @@ static void Main(string[] args)
             string assetsPath = GetAbsolutePath(AssetsRelativePath);
             string zipDataSet = Path.Combine(assetsPath, "input", "creditcardfraud-dataset.zip");
             string fullDataSetFilePath = Path.Combine(assetsPath, "input", "creditcard.csv");
-            string trainDataSetFilePath = Path.Combine(assetsPath, "output", "testData.csv");
-            string testDataSetFilePath = Path.Combine(assetsPath, "output", "trainData.csv");
+            string trainDataSetFilePath = Path.Combine(assetsPath, "output", "trainData.csv");
```
prathyusha12345: Thanks for identifying this. But better if we separate c# and f# changes into different PR's.

I fixed it in master.

prathyusha12345:
I see that we get the same output with c# samples for CreditCardFraudDetection. I have created Issue https://github.com/dotnet/machinelearning-samples/issues/306.

Other than above all samples are working fine. Thanks for the changes. Great job 👍

Issue #227, opened 2019-02-03
Format exception in BikeSharingDemand (Use culture when parsing)


bukaOK:
ReadSampleDataFromCsvFile method may cause Format exception if current culture doesn't support floating point numbers

CESARDELATORRE:
Thanks for catching this.

https://github.com/dotnet/machinelearning-samples/blob/b6455b2d48f54d72f3ae530ea827eed272edd838/samples/csharp/getting-started/Regression_BikeSharingDemand/BikeSharingDemand/BikeSharingDemandConsoleApp/ModelScoringTester.cs#L46-L58

                    Season = float.Parse(x[2]),
                    Year = float.Parse(x[3]),
                    Month = float.Parse(x[4]),
                    Hour = float.Parse(x[5]),
                    Holiday = float.Parse(x[6]),
                    Weekday = float.Parse(x[7]),
                    WorkingDay = float.Parse(x[8]),
                    Weather = float.Parse(x[9]),
                    Temperature = float.Parse(x[10]),
                    NormalizedTemperature = float.Parse(x[11]),
                    Humidity = float.Parse(x[12]),
                    Windspeed = float.Parse(x[13]),
                    Count = float.Parse(x[16])

Those should be changed to:

float.Parse(x[i], CultureInfo.InvariantCulture);

When the current culture is a culture that has a comma as a decimal separator, float.Parse will throw when the string uses a period `.`. Just like you’d imagine it to throw if it came across “23,4”.

The fix is to specify which culture to use to parse the strings.

We'll fix this in upcoming updates in the samples.
Thanks! 👍

prathyusha12345:
@CESARDELATORRE  I fixed this.So I am closing this issue

PR #312, opened 2019-03-15
Use culture when parsing a string to a float32

tikrimi:
It's the same fix than for issue https://github.com/dotnet/machinelearning-samples/issues/227 but on F# side.


prathyusha12345:
Thanks for changes. I am merging the PR. 👍

PR #328, opened 2019-03-23
Replace TaxiFarePrediction README's use of `ReadFromTextFile` with `LoadFromTextFile`

rdavisau:
This change is consistent with the actual sample code - I guess there was an API change at some point renaming the method from `ReadFromTextFile` to `LoadFromTextFile`.


prathyusha12345:
@rdavisau Thanks for identifying and fixing this.  I am merging this PR. Thanks 👍

PR #338, opened 2019-03-28
Normilizaing data for Spam prediction for 0.11.0 version

### LINKED ISSUE #331: v.11. How to lower the threshold of the predictor
In version 0.10 we were able to adjust the threshold for StochasticDualCoordinateAscent
[here](https://github.com/dotnet/machinelearning-samples/blob/73803ce861e16b48edefaad97771bcd721e268d6/samples/csharp/getting-started/BinaryClassification_SpamDetection/SpamDetectionConsoleApp/Program.cs#L61)

```
// The dataset we have is skewed, as there are many more non-spam messages than spam messages.
            // While our model is relatively good at detecting the difference, this skewness leads it to always
            // say the message is not spam. We deal with this by lowering the threshold of the predictor. In reality,
            // it is useful to look at the precision-recall curve to identify the best possible threshold.
            var inPipe = new TransformerChain<ITransformer>(model.Take(model.Count() - 1).ToArray());
            var lastTransformer = new BinaryPredictionTransformer<IPredictorProducing<float>>(mlContext, model.LastTransformer.Model, inPipe.GetOutputSchema(data.Schema), model.LastTransformer.FeatureColumn, threshold: 0.15f, thresholdColumn: DefaultColumnNames.Probability);

            ITransformer[] parts = model.ToArray();
            parts[parts.Length - 1] = lastTransformer;
            ITransformer newModel = new TransformerChain<ITransformer>(parts);
```

- [Cannot produce a Precision-Recall Curve from our Binary Classification Evaluator](https://github.com/dotnet/machinelearning/issues/2645)
- [Cannot set the threshold on a binary predictor](https://github.com/dotnet/machinelearning/issues/2465)

kdcllc:
Resolves #331 .  @CESARDELATORRE please let me know if this solution will work for now.


samples/csharp/getting-started/BinaryClassification_SpamDetection/SpamDetectionConsoleApp/Program.cs
```diff
@@ -36,41 +35,36 @@ static void Main(string[] args)
...
-                .Append(mlContext.Transforms.Text.FeaturizeText(outputColumnName: DefaultColumnNames.Features, inputColumnName: nameof(SpamInput.Message)));
-            
+            var dataProcessPipeLine = mlContext.Transforms.CustomMapping<InputRow, OutputRow>(CustomMappings.IncomeMapping, nameof(CustomMappings.IncomeMapping))
+                .Append(mlContext.Transforms.Text.FeaturizeText(outputColumnName: DefaultColumnNames.Features, inputColumnName: nameof(SpamInput.Message)))
+                                // The dataset we have is skewed, as there are many more non-spam messages than spam messages.
+                                // While our model is relatively good at detecting the difference, this skewness leads it to always
+                                // say the message is not spam. We deal with this by lowering the threshold of the predictor. In reality,
+                                // it is useful to look at the precision-recall curve to identify the best possible threshold.
+                .Append(mlContext.Transforms.Normalize(new NormalizingEstimator.LogMeanVarColumnOptions("LogMeanVarNormalized", DefaultColumnNames.Features, useCdf: true)))
```
prathyusha12345: @kdcllc  Could you  please explain what normalize function does here and explain about the parameters?

samples/csharp/getting-started/BinaryClassification_SpamDetection/SpamDetectionConsoleApp/Program.cs
```diff
@@ -36,41 +35,36 @@ static void Main(string[] args)
...
 
             // Now let's train a model on the full dataset to help us get better results
             var model = trainingPipeLine.Fit(data);
 
-            // The dataset we have is skewed, as there are many more non-spam messages than spam messages.
-            // While our model is relatively good at detecting the difference, this skewness leads it to always
-            // say the message is not spam. We deal with this by lowering the threshold of the predictor. In reality,
-            // it is useful to look at the precision-recall curve to identify the best possible threshold.
-            var inPipe = new TransformerChain<ITransformer>(model.Take(model.Count() - 1).ToArray());
-            var lastTransformer = new BinaryPredictionTransformer<IPredictorProducing<float>>(mlContext, model.LastTransformer.Model, inPipe.GetOutputSchema(data.Schema), model.LastTransformer.FeatureColumn, threshold: 0.15f, thresholdColumn: DefaultColumnNames.Probability);
```
justinormont: Were you looking to remove the `threshold: 0.15f`? If there's an ok reason to, you'll want to also remove the comment which discusses the dataset skew: `The dataset we have is skewed, as there are many m...`.

samples/csharp/getting-started/BinaryClassification_SpamDetection/SpamDetectionConsoleApp/Program.cs
justinormont: Though it can help in certain cases, you'll likely not want to renormalize the text column. The output of the `FeaturizeText()` runs though a LpNorm with defaults of L2 normalization. 

The row-wise normalization from the LpNorm and the column-wise LogMeanVar normalization can interact poorly leading to worse results on many datasets. I'd recommend not adding it.

justinormont: I might recommend a pipeline of:
```C#
// Data process configuration with pipeline data transformations 
var dataProcessPipeline = mlContext.Transforms.Conversion.MapValueToKey("Label", "Label")
                          .Append(mlContext.Transforms.Text.FeaturizeText("FeaturesText", new Microsoft.ML.Transforms.Text.TextFeaturizingEstimator.Options
                          {
                              WordFeatureExtractor = new Microsoft.ML.Transforms.Text.WordBagEstimator.Options { NgramLength = 2, UseAllLengths = true },
                              CharFeatureExtractor = new Microsoft.ML.Transforms.Text.WordBagEstimator.Options { NgramLength = 3, UseAllLengths = false },
                          }, "text"))
                          .Append(mlContext.Transforms.CopyColumns("Features", "FeaturesText"))
                          .Append(mlContext.Transforms.Normalize("Features", "Features"))
                          .AppendCacheCheckpoint(mlContext);

// Set the training algorithm 
var trainer = mlContext.MulticlassClassification.Trainers.OneVersusAll(mlContext.BinaryClassification.Trainers.AveragedPerceptron(labelColumnName: "Label", numberOfIterations: 10, featureColumnName: "Features"), labelColumnName: "Label")
                          .Append(mlContext.Transforms.Conversion.MapKeyToValue("PredictedLabel", "PredictedLabel"));
var trainingPipeline = dataProcessPipeline.Append(trainer);

return trainingPipeline;
```

Scores from this pipeline:
```md
*************************************************************************************************************
*       Metrics for Multi-class Classification model      
*------------------------------------------------------------------------------------------------------------
*       Average MicroAccuracy:    0.9899  - Standard deviation: (0.0050)  - Confidence Interval 95%: (0.0033)
*       Average MacroAccuracy:    0.9653  - Standard deviation: (0.0154)  - Confidence Interval 95%: (0.0101)
*       Average LogLoss:          0.0391  - Standard deviation: (0.0144)  - Confidence Interval 95%: (0.0094)
*       Average LogLossReduction: 0.9012  - Standard deviation: (0.0328)  - Confidence Interval 95%: (0.0214)
*************************************************************************************************************
```

To more easily handle the text labels, I'm running as multi-class. 

The label skew doesn't seem to have a large effect. The non-spam class is 87% of the dataset, and at around 99% accuracy, the model is well above just guessing the largest class.

kdcllc: > Though it can help in certain cases, you'll likely not want to renormalize the text column. The output of the `FeaturizeText()` runs though a LpNorm with defaults of L2 normalization.
> 
> The row-wise normalization from the LpNorm and the column-wise LogMeanVar normalization can interact poorly leading to worse results on many datasets. I'd recommend not adding it.

Since in version 0.11.0 there is not way to adjust threshold. I was attempting to `renormalize` some of the data in the dataset. I am no Data Scientist. This is a good explanation of what I thought it might cause with larger dataset. Thank you.

samples/csharp/getting-started/BinaryClassification_SpamDetection/SpamDetectionConsoleApp/Program.cs
kdcllc: @justinormont I was attempting to denormilize the dataset  somehow to allow for the balance in the predictions. If you are okay, I was trying to still use `StochasticDualCoordinateAscent` as the original sample did. I can update the PR with your proposed pipeline. What do you think?

justinormont: There's no specific reason to use `SDCA`. `AveragedPerceptron` does better on most text datasets; I would recommend it.

The normalization that you're using won't interact with the Label Skew. In addition, the Label Skew is rather minor in this dataset (~7:1). Generally, I start to look into the skew around 10:1, and act around 20:1.

Choosing the best pipeline for an example is slightly more involved that just what's best for the given dataset as sample code is designed to demonstrate functionality and to teach best practices. 

I would purpose using my [pipeline](https://github.com/dotnet/machinelearning-samples/pull/338#discussion_r270294570), and adding another sample which actually has need of correcting the Label Skew, or has a strong business case for weighting false-positive and false-negative differently. Perhaps the Wisconsin Breast Cancer dataset, where a false-negative is significantly worse then a false-positive. Another route, would be to move the dataset used in this example.

prathyusha12345:
@kdcllc  Thanks for your changes. However I have migrated the sample to ML.Net 1.0.0-preview with pipeLine recommended by @justinormont as it gives better accuracy metrics. Please find the sample here https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/BinaryClassification_SpamDetection

In future we will add a sample based on label skew. I will close this PR. Thanks

Adding @CESARDELATORRE  for reference.

PR #341, opened 2019-04-03
C# create Object Detection sample using Onnx Model

prathyusha12345:
-  Created the Object Detection sample. This sample uses TIny_Yolo2 Onnx model.

- The version of the packages are as below. We don't have 1.1.0-preview version for  **Microsoft.ML.OnnxTransformer** package

  **Microsoft.ML -  1.0.0-preview
    Microsoft.ML.ImageAnalytics - 1.0.0-preview
    Microsoft.ML.OnnxTransformer - 0.12.0-preview**

The o/p of the sample is given below.

image


jignparm:
Please add a README.md file, similar to https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/MulticlassClassification_Iris. This makes it easier for users to see what the demo is about.

samples/csharp/getting-started/DeepLearning_ObjectDetection_Onnx/ObjectDetectionConsoleApp/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public static string GetAbsolutePath(string relativePath)
+        {
+            FileInfo _dataRoot = new FileInfo(typeof(Program).Assembly.Location);
+            string assemblyFolderPath = _dataRoot.Directory.FullName;
+
+            string fullPath = Path.Combine(assemblyFolderPath, relativePath);
+
+            return fullPath;
```
jignparm: we can probably remove the spaces.

prathyusha12345: ok..As part of adding ReadMe fileCould you provide me the name of algorithm that is used to train the Tiny_yolo2 model?

samples/csharp/getting-started/DeepLearning_ObjectDetection_Onnx/ObjectDetectionConsoleApp/OnnxModelScorer.cs
```diff
@@ -0,0 +1,100 @@
...
+            this.imagesFolder = imagesFolder;
+            this.modelLocation = modelLocation;
+            mlContext = new MLContext();
+        }
+
+        public struct ImageNetSettings
+        {
+            public const int imageHeight = 416;
+            public const int imageWidth = 416;
+            public const bool channelsLast = true;
```
jignparm: Remove -- this is not used.

samples/csharp/getting-started/DeepLearning_ObjectDetection_Onnx/ObjectDetectionConsoleApp/OnnxModelScorer.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            return predictionEngine;
+        }
+
+        protected void PredictDataUsingModel(string imagesLocation,
+                                                                  string imagesFolder,
+                                                                  PredictionEngine<ImageNetData, ImageNetPrediction> model)
+        {
+            Console.WriteLine("Identify the objects in the images");
+            Console.WriteLine("");
```
jignparm: Instead of printing comments here, is it easy to return the predictions values to the main() function and add all comments there? It makes the example more readable.

prathyusha12345: That's good approach. But currently I have followed the existing samples. For Ex: TensorFlow_Imagescorer is printing the output in the prediction method. I think we need to change it for all the samples. 

@CESARDELATORRE Do you also suggest to change this?

samples/csharp/getting-started/DeepLearning_ObjectDetection_Onnx/ObjectDetectionConsoleApp/OnnxModelScorer.cs
prathyusha12345: Ok. I have removed ChannelsLast variable.

jignparm:
Instead of checking in the TinyYoloV2.onnx model, can we just refer to the model from the onnx model zoo? Or download it on the fly in the project? Large binaries are typically not checked into GitHub.

chris-lauren:
Download on the fly would be great … general “get model” method would be pretty helpful

prathyusha12345:
Ok I will modify the code to download the model from the url after merging this PR. I am working on that code and push another PR

Issue #345, opened 2019-04-04
String to float conversion


PeterSoder:
String to float conversion does not work in cultures with something other then period as decimal separator, maybe use CultureInfo.InvariantCulture explicitly.

https://github.com/dotnet/machinelearning-samples/blob/cb6ba8437bfc7175d2815a378510f88f115cecd7/samples/csharp/getting-started/Regression_TaxiFarePrediction/TaxiFarePrediction/TaxiFarePredictionConsoleApp/Program.cs#L317

prathyusha12345:
As this issue is fixed, I am closing the issue.

PR #357, opened 2019-04-09
Parse float from string with invariant culture

tikrimi:
This PR is to fix the issue [#345](https://github.com/dotnet/machinelearning-samples/issues/345).


prathyusha12345:
@tikrimi Thanks for your changes. However we prefer the code should be added per sample instead of adding the code in common location. In future we are going to have just **ConsoleHelper.cs** file in **common** folder. Our goal is every sample should be autonomous. 

So Could you please remove the **ParsingExtensions** class in **common** folder and add the code in all the required samples. 

Adding @CESARDELATORRE  for reference.

tikrimi:
Ok, it makes sense. So I will use float.Parse(val, CultureInfo.InvariantCulture) where it's needed.

tikrimi:
I updated my PR. Should be ok now.

tikrimi:
Arf, I have a conflic due to alphabetic sorting of "using". Will fix it.

prathyusha12345:
@tikrimi  Thanks for the changes. I tested the samples and they are working fine. I am merging the changes. Nice job 👍

PR #366, opened 2019-04-12
F# update spam sample method

kevmal:


prathyusha12345:
Great!!! Thanks for doing changes...

PR #370, opened 2019-04-15
Fix type-o in readme: 'speccifying' => 'specifying'

tpayne84:
Fix type-o in readme: 'speccifying' => 'specifying'.


prathyusha12345:
Thanks for fixing. I am merging PR. Thanks 👍

Issue #368, opened 2019-04-13
Question regarding MNIST example


jeffprosise:
The MNIST example works. But how do you correlate the probabilities in the Score array in the output to the labels? If I pass the pattern for a 4 to Predict(), it is the element at index 2 in the Score array that shows the highest probability.

kevmal:
The example maps the actual number to a key by `mlContext.Transforms.Conversion.MapValueToKey("Label", "Number")`, the default here is to assign keys to values by occurrence. So in this case 4 corresponds to index 2 because two distinct values appear in the data before the first 4.  To get a more intuitive ordering the optional parameter `keyOrdinality` can be set to specify `ByValue` ordering.

prathyusha12345:
Fixed this. so closing the issue. Please reopen if you still have any issues.

PR #377, opened 2019-04-18
Fix key ordering in mnist sample

kevmal:
This addresses issue #368. By default `MapValueToKey` uses first occurrence ordering of keys where the sample output is assuming value ordering.


prathyusha12345:
Thanks for addressing the issue. I tested the PR and it works fine. I am merging the changes. Thanks 👍

PR #385, opened 2019-04-22
F# heart disease detection sample

kevmal:


prathyusha12345:
@kevmal Thanks for the sample. I tested the sample and it works fine. I am merging the PR. Thanks 👍

PR #387, opened 2019-04-24
Rename shampoo sales to  AnomalyDetection_sales

prathyusha12345:
Renamed shampoo sales to generic product sales so that this sample will be suitable for all kinds of product sales data


prathyusha12345:
I changed the names in ReadMe file also. Everything works fine. I am merging the PR

PR #391, opened 2019-04-25
Remove all files in Common folder except ConsoleHelper

prathyusha12345:
-Removed ModelHelper and objectPool.cs classes and added them in a new library project in salesForecast solution. Referenced this library in eshopDashBoard project and objectpoolingConsoleApp project.
-Removed Model Builder.cs , ModelScorer.cs classes as they are not used at all.


prathyusha12345:
I am merging as this is part of cleanup of classes from common folder. the sample works fine.

Issue #398, opened 2019-05-01
Regression-SalesForecast no longer builds successfully.


montebhoover:
[Regression-SalesForecast](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/end-to-end-apps/Regression-SalesForecast) no longer builds after cloning from master due to the refactoring in #391.  A new project "CommonHelpers" was added, but it looks like it doesn't build with projects that reference it.  If CommonHelpers.csproj is explicitly added to the .sln it will build properly.

montebhoover:
By the way, the SalesForecast web app sample is really impressive.  Well done!

prathyusha12345:
@montebhoover I fixed the issue and pushed into master. We need to add reference to commonHelpers.csproj in solution to fix the issue.Could you please take latest of master and try again.

montebhoover:
Looks good, thanks!

PR #400, opened 2019-05-01
Added CommonHelpers.csproj to Regression-SalesForecast .sln

montebhoover:
Regression-SalesForecast doesn't build out of the box right now due to #398 where CommonHelpers.csproj doesn't build along with the projects that reference it.  This PR fixes that by explicitly adding it to the .sln.


prathyusha12345:
I fixed this in master before looking at this PR. Thanks for the PR. I am closing this PR as its fixed . Thanks 👍

Issue #399, opened 2019-05-01
Clarify a few things in Regression-SalesForecast documentation.


montebhoover:
In #397 I had a problem with the web view because it wasn't apparent to me that I needed to enter a search term first in order to get the web page to populate.  Maybe adding a line stating so would help users like me - I'll let you decide.  Also, the first portion of the documentation deals with NuGet configuration but no action is necessary (only optional), so maybe it would help to get straight to the actionable steps and mention the NuGet configs as a footnote or something.  I'll create a PR with some suggested edits and feel free to take as much or as little as you find helpful.

prathyusha12345:
Great!!! Thanks for coming up with suggestion.It would be better to add the info in ReadMe file. Let me add it and update.

prathyusha12345:
I have added the info in ReadMe file in this commit https://github.com/dotnet/machinelearning-samples/commit/dbe1a9cad04b4285a8be8b2dd95f0bb35ed8e45a#diff-2fb36f80bcbc39cf9c2bff427bce6fa4. 

I am closing the issue. Feel free to reopen this issue if you still need any more info to be added.

Issue #397, opened 2019-04-29
Web view not populating in Regression-SalesForecast


montebhoover:
In [Regression-SalesForecast](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/end-to-end-apps/Regression-SalesForecast) the website does not seem to be working out of the box.  The web page serves, but doesn't get populated with info as shown in the readme.  Here's what it looked like when I ran it from the repo with no modifications:

![image](https://user-images.githubusercontent.com/37886197/56904025-1e7af200-6aa6-11e9-8c86-b3c56c914ed3.png)

Repro steps:
1. Clone dotnet/machinelearning-samples
2. Open eShopDashboardML.sln in Visual Studio 2019
3. Press F5 to run eShopDashboard project in Debug configuration
4. Web page displays , but no content is populated

prathyusha12345:
You need to search for a product with say "bottle" in search box where **Type a product** tool tip string displays like below.
![image](https://user-images.githubusercontent.com/22335043/56922749-d1832580-6a7d-11e9-92e3-21f4d64c8dfa.png)

once you select any product from search drop down, and press Enter, it will show the product forecast results as shown below.

![image](https://user-images.githubusercontent.com/22335043/56922875-127b3a00-6a7e-11e9-8437-b140bda55911.png)

prathyusha12345:
I fixed one build issue in master. Please take latest changes in your repo.

montebhoover:
Thank you, this looks great.  Thanks for adding the FastTree reference into master.  The instruction to search for a product is exactly what I needed - if you don't mind I'll create a PR to add that to the instructions and you can decide if it is helpful.

Also, in one of the recent changes the CommonHelpers csproj was added, but it looks like it still needs to get added to the .sln to work out of the box.  I created #398 to make note of this.  Also, I created #399 to mention a few documentation suggestions in case you find them helpful.

Thanks for making this great sample!

PR #401, opened 2019-05-01
Edits to Regression-SalesForecast docs regarding the web view.

montebhoover:
I had a little trouble understanding how to use the eShopDashboard as noted in #397 and #399, so I clarified the docs in a way that would have helped me.  Feel free to take some or none of this change if you find it helpful.


prathyusha12345:
Great!!! Thanks for the PR. However I have added the info in ReadMe file regarding searching for a product when the app starts in this commit https://github.com/dotnet/machinelearning-samples/commit/dbe1a9cad04b4285a8be8b2dd95f0bb35ed8e45a

I think its better to keeping the current structure instead of creating nuget-configuration.md file. So I am closing this PR. Thanks 👍

PR #416, opened 2019-05-06
Create webapp for Live Object Detection Demo

prathyusha12345:
Created the webapp to select the image from the list. 

when the app starts

image

when you click on any image, the object are being drawn with bounding boxes on the image and the image is shown like below.

image


prathyusha12345:
Closing this PR as I already added this to end-to-end samples by adding a new branch from master.

Issue #458, opened 2019-05-15
AutoML multiclass (MNIST) sample predictions


kevmal:
In `TestSomePredictions` the output states "Actual: 7" when the sample corresponds to a 1 and "Actual 1" when the data used corresponds to 7.

kevmal:
#459 fixed this

PR #459, opened 2019-05-15
MNIST Auto ML - Add prefeaturizer for predictable Score[] ordering

kevmal:
Addresses #458

### DIFF
--- a/samples/csharp/getting-started/MulticlassClassification_AutoML/MNIST/Program.cs
+++ b/samples/csharp/getting-started/MulticlassClassification_AutoML/MNIST/Program.cs
@@ -65,9 +65,10 @@ namespace MNIST
                 // STEP 3: Run an AutoML multiclass classification experiment
                 ConsoleHelper.ConsoleWriteHeader("=============== Running AutoML experiment ===============");
                 Console.WriteLine($"Running AutoML multiclass classification experiment for {ExperimentTime} seconds...");
+                var preFeaturizer = mlContext.Transforms.Conversion.MapValueToKey("Number", "Number", keyOrdinality: Microsoft.ML.Transforms.ValueToKeyMappingEstimator.KeyOrdinality.ByValue);
                 ExperimentResult<MulticlassClassificationMetrics> experimentResult = mlContext.Auto()
                     .CreateMulticlassClassificationExperiment(ExperimentTime)
-                    .Execute(trainData, "Number", progressHandler: progressHandler);
+                    .Execute(trainData, "Number", progressHandler: progressHandler, preFeaturizer: preFeaturizer);
 
                 // Print top models found by AutoML
                 Console.WriteLine();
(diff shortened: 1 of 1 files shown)


srsaggam:
@CESARDELATORRE @justinormont

samples/csharp/getting-started/MulticlassClassification_AutoML/MNIST/Program.cs
```diff
@@ -65,9 +65,10 @@ public static void Train(MLContext mlContext)
                 // STEP 3: Run an AutoML multiclass classification experiment
                 ConsoleHelper.ConsoleWriteHeader("=============== Running AutoML experiment ===============");
                 Console.WriteLine($"Running AutoML multiclass classification experiment for {ExperimentTime} seconds...");
+                var preFeaturizer = mlContext.Transforms.Conversion.MapValueToKey("Number", "Number", keyOrdinality: Microsoft.ML.Transforms.ValueToKeyMappingEstimator.KeyOrdinality.ByValue);
```
justinormont: I don't think you need the sort order specified. But you'll want a `MapKeyToValue` at the end. These should be automatically generated for you by the AutoML.

kevmal: The default is by occurrence if the parameter is unspecified. AutoML does automatically generate the key mappings however it too is by occurrence. Confirmed by opening the Model.zip generated by the master branch and checking `Vocabulary/Terms.txt` of the transform
```
# Number of terms of type 'Single' = 10
0	0
1	7
2	4
3	6
4	2
5	5
6	8
7	1
8	9
9	3
```
Which looking in `optdigits-train.csv` corresponds to order of appearance. Adding the prefeaturizer it seems AutoMl still generates the `MapKeyToValue` at the end (confirmed by `LastEstimator` field on the best model), though it's an assumption on my part that AutoML will not "rekey" a key column and that `MapKeyToValue`  corresponds to the prefeaturizer transform.

CESARDELATORRE: @justinormont - Is this PR about AutoML API tested and ready to be merged?

justinormont: @daholste: Do we automatically append a `MapKeyToValue()` at the end of the AutoML pipeline to produce the original Label values (even with numeric inputs)? If not, we likely should be.

daholste: yeah, we do append MapKeyToValue() at the end of the AutoML pipeline

justinormont: Since we add the `MapKeyToValue()` to the pipeline, the original values should be printed. 

Adding the `KeyOrdinality.ByValue` is an interesting method to ensure the internal representation also matches the file. Though this method only works specifically for this dataset as the classes are numeric and start at 0 and are not missing any values. 

If the class values aren't matching after the `MapKeyToValue()`, we should investigate that.
Basically, the `MapKeyToValue()` is the right way to obtain the original class values back. The purposed method will not work for the general case, so I wouldn't put it in a sample.

kevmal:
`MapKeyToValue()` is fine since `PredictedLabel` is actually correct. Problem is this sample is outputting all the scores with the assumption `Score[index]` corresponds to number `index`. The point that using `ByValue` is a bit specific to this sample is a good one. 

Updated to instead use the key-value mapping from the output schema.

kevmal:
@justinormont Any other thoughts on this? Using `GetKeyValues` seems to be the way to do this from: https://github.com/dotnet/machinelearning/issues/1978

justinormont:
@wschin, @artidoro: is `GetKeyValues()`, as seen in this PR, the correct method of matching the `Score[]` values to their original class names?

For instance, if I wanted to create an autosuggest, I would need to get the top ~3 classes for an instance. Then I would need to get the class names of the top 3 classes. Is `GetKeyValues()` the right method to get the class names?

bamurtaugh:
@wschin, @artidoro, @justinormont - Any further updates/confirmation about this PR?

PR #470, opened 2019-05-20
Make Sentiment Analysis sample work with Try .NET

colbylwilliams:
Simplified sample code and updated README to work with [Try .NET](https://github.com/dotnet/try)


CESARDELATORRE:
This PR is nice! For now we can keep this sample compatible with "Try .NET".
There are other competing strategies like 'Jupyter notebooks' with .NET, but that is not ready, yet. 

So let's have this sample with "Try .NET" and we'll see in the future.

Thanks a lot! 👍

CESARDELATORRE:
And btw, we might use TRY .NET for the ML.NET marketing pages because TRY .NET is perfect for that scenario, like here: 👍 

https://dotnet.microsoft.com/apps/machinelearning-ai/ml-dotnet

PR #480, opened 2019-05-23
Log Exception.ToString() instead of Exception.Message

AkosLukacs:
Hi!

Thanks for the samples! The first thing I ran into, is only the exception's `Message` is logged or written to console. But that's not always enough. InnerException and Stacktrace helps a lot. For example:

```
EXCEPTION
#########
Splitter/consolidator worker encountered exception while consuming source data
```
There was a typo in the filename:
```
EXCEPTION
#########
System.InvalidOperationException: Splitter/consolidator worker encountered exception while consuming source data
 ---> System.InvalidOperationException: Image xtoaster2.png was not found.
```

Or 

```
EXCEPTION
#########
Index was outside the bounds of the array.
```
Stacktrace helps, checked the code at that line, and found the reason:
```
EXCEPTION
#########
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Program.classifyImages@121-23.Invoke(String x) in ........\Program.fs:line 121
```

I think the simplest thing is to use `ToString()`, that usually provides enough details.
Sorry about the white space changes. Are you OK with it as-is?

### DIFF
--- a/samples/csharp/end-to-end-apps/Regression-SalesForecast/src/eShopForecastModelsTrainer/Program.cs
+++ b/samples/csharp/end-to-end-apps/Regression-SalesForecast/src/eShopForecastModelsTrainer/Program.cs
@@ -17,11 +17,11 @@ namespace eShopForecastModelsTrainer
 
         static void Main(string[] args)
         {
-            
+
             try
             {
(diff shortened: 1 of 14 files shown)


prathyusha12345:
I think this is better idea showing full exception details. I will remove the white spaces you have added. I am merging the PR . Thanks 👍

PR #517, opened 2019-06-12
Fixed GetImagesData to only enumerate the jpg files

harishsk:
Fixes #3675

This bug was actually reported for the Mac. 

On the Mac, this specific bug doesn't repro if you make sure libgdiplus is installed and upgraded to the latest version. 

However, on both Mac and Windows, there is still a crash as the sample code tries to enumerate and run the prediction on all the files in the assets directory. That assets directory also has a wikimedia.md file which is not an image file. The ImageLoadingTransformer tries to load the wikimedia.md file as a bitmap which leads to a crash in System.Drawing.Bitmap constructor with a System.ArgumentException.

So, I have changed GetImagesData function to only enumerate the jpg files in the assets directory. 

This should fix the crash on both Windows and Mac.


harishsk:
My fix wasn't necessary as the refactoring done a couple of hours ago has fixed the issue.

PR #533, opened 2019-06-25
Ranking Sample - Hotel Search Results (changed to Bing Search Engine result ranking)

nicolehaugen:
Please review the sample that I have created to show how to do ranking using Light GBM.
Note that I have changed this sample from using the Expedia dataset to instead use Bing's search result dataset.


samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Common/ConsoleHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+            Console.WriteLine($"NDCG: {string.Join(", ", metrics.NormalizedDiscountedCumulativeGains.Select((d, i) => $"@{i + 1}:{d:F4}").ToArray())}");
+        }
+
+        // Performs evaluation with the truncation level set up to 10 hotel search results within a query.
+        // This is a temporary workaround for this issue: https://github.com/dotnet/machinelearning/issues/2728.
+        public static void EvaluateMetrics(MLContext mlContext, IDataView scoredData, int truncationLevel)
+        {
+            if (truncationLevel < 1 || truncationLevel > 10)
+            {
+                throw new InvalidOperationException("Currently metrics are only supported for 1 to 10 truncation levels.");
```
justinormont: Is there something causing only support up to n=10?

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                Console.WriteLine(e.Message);
+            }
+
+            Console.ReadLine();
+        }
+
+        static void PrepDatasets(MLContext mlContext, string assetPath, string originalDatasetPath, string trainDatasetPath, string testDatasetPath)
+        {
+            const string DatasetUrl = "https://www.kaggle.com/c/expedia-personalized-sort/download/data.zip";
```
justinormont: File is not public. Is requiring a user to sign up for a kaggle account too onerous? 

I'd recommend using the MSLR-WEB10K (or WEB30K) dataset from our CDN:
* https://aka.ms/mlnet-resources/benchmarks/MSLRWeb10KTrain720kRows.tsv
* https://aka.ms/mlnet-resources/benchmarks/MSLRWeb10KValidate240kRows.tsv
* https://aka.ms/mlnet-resources/benchmarks/MSLRWeb10KTest240kRows.tsv

As zip:
* MSLR-WEB10K: <br>https://express-tlcresources.azureedge.net/datasets/MSLR-WEB10K/MSLR-WEB10K.zip ([README](https://express-tlcresources.azureedge.net/datasets/MSLR-WEB10K/README.md), [LICENSE](https://express-tlcresources.azureedge.net/datasets/MSLR-WEB10K/LICENSE.md))
* MSLR-WEB30K: <br>https://express-tlcresources.azureedge.net/datasets/MSLR-WEB30K/MSLR-WEB30K.zip ([README](https://express-tlcresources.azureedge.net/datasets/MSLR-WEB30K/README.md), [LICENSE](https://express-tlcresources.azureedge.net/datasets/MSLR-WEB30K/LICENSE.md))

We use the MSLR-WEB10K dataset for our [ranking benchmarks](https://github.com/dotnet/machinelearning/blob/08318656ed0f0649aa75370b019cf4bcbda5a6a5/test/Microsoft.ML.Benchmarks/Numeric/Ranking.cs). The dataset is auto-downloaded from the above CDN aka.ms URLs when you run `build.cmd -- /t:DownloadExternalTestFiles /p:IncludeBenchmarkData=true` to [download/run](https://github.com/dotnet/machinelearning/blob/8444a5a2a4c6b162c35b98cbff95983db3d676c2/test/Microsoft.ML.Benchmarks/README.md#run-the-performance-tests) the benchmark datasets.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/DataStructures/HotelPrediction.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        // Maps to the "Srch_Id" column in the underlying Expedia dataset; this is the id of the search\query.
+        public uint GroupId { get; set; }
+
+        // The "Label" does not exist in the underlying Expedia dataset and is added in the sample to indicate the ideal rank of a hotel search result.  This is 2 if the user purchased\booked a room at this hotel.  Or, is 1 if the user clicked through to see more information on this hotel.  Otherwise, is 0 if the user neither clicked nor purchased\booked a room at this hotel. 
+        public uint Label { get; set; }
+
+        // Prediction made by the model that indicates the relative ranking of the hotel search result.
+        [ColumnName("Score")]
+        public float PredictedRank { get; set; }
```
justinormont: Terminology -- I don't consider the term `score` to be the same as `rank`:
```suggestion
        public float Score { get; set; }
```

The result gets ranked first after you sort your list by the score. At that point the result is ranked first in the results.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
```diff
@@ -0,0 +1,189 @@
...
+                    // To transform the data, call the Fit() method.
+                    ITransformer dataTransformer = dataPipeline.Fit(data);
+                    IDataView labeledData = dataTransformer.Transform(data);
+
+                    Console.WriteLine("===== Split the data into testing/training datasets =====");
+
+                    // When splitting the data, 20% is held for the test dataset.
+                    // To avoid label leakage, the GroupId (e.g. search\query id) is specified as the samplingKeyColumnName.  
+                    // This ensures that if two or more hotel instances share the same GroupId, that they are guaranteed to appear in the same subset of data (train or test).
+                    TrainTestData trainTestData = mlContext.Data.TrainTestSplit(labeledData, testFraction: .2, samplingKeyColumnName: nameof(HotelData.GroupId), seed: 1);
```
justinormont: Leading zero:
```suggestion
                    TrainTestData trainTestData = mlContext.Data.TrainTestSplit(labeledData, testFraction: 0.2, samplingKeyColumnName: nameof(HotelData.GroupId), seed: 1);
```

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            // Predict rankings.
+            IDataView predictions = predictionPipeline.Transform(exampleData);
+
+            // In the predictions, get the scores of the hotel search results included in the first query (e.g. group).
+            IEnumerable<HotelPrediction> hotelQueries = mlContext.Data.CreateEnumerable<HotelPrediction>(predictions, reuseRowObject: false);
+            var firstGroupId = hotelQueries.First<HotelPrediction>().GroupId;
+            IEnumerable<HotelPrediction> firstGroupPredictions = hotelQueries.Take(50).Where(p => p.GroupId == firstGroupId).OrderByDescending(p => p.PredictedRank).ToList();
+
+            // The individual scores themselves are NOT a useful measure of accuracy; instead, they are used to determine the ranking where a higher score indicates a higher ranking.
```
justinormont: I'd avoid the term `accuracy` when discussing metrics, unless you're specifically talking about the `accuracy metric`:
```suggestion
            // The individual scores themselves are NOT a useful measure of result quality; they are only useful as a relative measure to other scores in the group. The scores are used to determine the ranking where a higher score indicates a higher ranking vs another result.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
+# Rank hotel search results to provide personalized sorting
+
+| ML.NET version | API type          | Status                        | App Type    | Data type | Scenario            | ML Task                   | Algorithms                  |
+|----------------|-------------------|-------------------------------|-------------|-----------|---------------------|---------------------------|-----------------------------|
+| v1.1.0         | Dynamic API       | Up-to-date                    | Console app | .csv file | Ranking hotel search results | Ranking          | LightGbm |
+
+This introductory sample shows how to use ML.NET to predict the relevance and order of hotel search results. In the world of machine learning, this type of prediction is known as ranking.
```
justinormont:
```suggestion
This introductory sample shows how to use ML.NET to predict the best order to display hotel search results. In the world of machine learning, this type of prediction is known as ranking.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+* Travel Agency - Provide a list of hotels with those that are most likely to be purchased\booked by the user positioned highest in the list.
+* Shopping - Display items from a product catalog in an order that aligns with a user's shopping preferences.
+* Recruiting - Retrieve job applications ranked according to the candidates that are most qualified for a new job opening.
+
+Ranking is useful to any scenario where it is important to list items in an order that increases the likelihood of a click, purchase, reservation, etc.
+ 
+In this sample, we show how to apply ranking to the first example listed above to rank hotel search results according to the likelihood that the hotel will be purchased\booked by the user. To perform ranking, there are two algorithms currently available - FastTree Boosting (FastRank) and Light Gradient Boosting (LightGbm).  We use the LightGbm's Lambdarank algorithm in this sample to automatically build an ML model to predict ranking.
```
justinormont:
```suggestion
In this sample, we show how to apply ranking to the first example listed above to rank hotel search results according to the likelihood that the hotel will be purchased\booked by the user. To perform ranking, there are two algorithms currently available - FastTree Boosting (FastRank) and Light Gradient Boosting Machine (LightGBM).  We use the LightGBM's LambdaRank implementation in this sample to automatically build an ML model to predict ranking. 
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+* Hotel attributes, such as location attractiveness and price.
+* User's criteria for searching hotels, such as the number of rooms\children\adults, length of stay, etc.
+* User's purchase and browsing history, such as whether they clicked the link of a hotel or purchased\booked it.
+* Information on similar competitor hotel offerings.
+
+## ML Task - Ranking
+As previously mentioned, this sample uses the LightGbm Lambdarank algorithm which is applied using a supervised learning technique known as "Learning to Rank".  This technique requires that train/test datasets contain groups of data instances that are labeled with their ideal ranking value.  The label is a numerical\ordinal value, such as {4, 3, 2, 1, 0} or a text value {"Perfect", "Excellent", "Good", "Fair", or "Bad"}.  The process for labeling these data instances with their ideal ranking value can be done manually by subject matter experts.  Or, the labels can be determined using other metrics, such as the number of clicks on a given search result. This sample uses the latter approach.
```
justinormont:
```suggestion
As previously mentioned, this sample uses the LightGBM LambdaRank algorithm which is applied using a supervised learning technique known as "[Learning to Rank](https://en.wikipedia.org/wiki/Learning_to_rank)".  This technique requires that train/test datasets contain groups of data instances that are each labeled with their relevance scores.  The label is a numerical\ordinal value, such as {4, 3, 2, 1, 0} or a text value {"Perfect", "Excellent", "Good", "Fair", or "Bad"}.  The process for labeling these data instances with their ideal ranking value can be done manually by subject matter experts.  Or, the labels can be determined using other metrics, such as the number of clicks on a given search result. This sample uses the latter approach.
```

justinormont: Has someone checked if `{"Perfect", "Excellent", "Good", "Fair", or "Bad"}` still work as Labels? This works from the MAML command-line, but I'm uncertain that the estimators API accepts anything besides {4, 3, 2, 1, 0}.

justinormont: We may also want to note that it's expected to have many more "Bad" relevance scores than "Perfect". This helps users to avoid converting a ranked list directly into a equally sized bins of {0, 1, 2, 3, 4}.

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+Once the train/test datasets are labeled with ideal ranking values, the model (e.g. ranker) can then be trained and tested using this data.  Through the model training process, the ranker learns how to score each data instance within a group based on their label value.  The resulting score of an individual data instance by itself isn't important -- instead, the scores should be compared against one another to determine the relative ordering of a group's data instances.  The higher the score a data instance has, the more relevant and more highly ranked it is within its group.      
+
+## Solution
+The sample performs the following high-level steps to rank Expedia hotel search results:
+1. Each hotel search result is **labeled** with its ideal ranking value.
```
justinormont:
```suggestion
1. Each hotel search result is **labeled** with its relevance score.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+2. Once the dataset is labeled, the data is **split** into training and testing datasets.  
+3. The model is **trained** using the train dataset using LightGbm Lambdarank algorithm.
```
justinormont:
```suggestion
3. The model is **trained** using the train dataset using LightGBM LambdaRank algorithm.  
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Solution
+The sample performs the following high-level steps to rank Expedia hotel search results:
+1. Each hotel search result is **labeled** with its ideal ranking value.
+2. Once the dataset is labeled, the data is **split** into training and testing datasets.  
+3. The model is **trained** using the train dataset using LightGbm Lambdarank algorithm.  
+4. The model is **tested** using the test dataset.  This results in a **prediction** that includes a **score** for each hotel instance.  The score is used to determine the ranking relative to other hotels within the same query (e.g. group).  The predictions are then **evaluated** by examining metrics; specifically the [Discounted Cumulative Gain](https://en.wikipedia.org/wiki/Discounted_cumulative_gain).
```
justinormont: I'd focus on the NDCG metric as its comparable across datasets, whereas DCG heavily depends on the dataset:
```suggestion
4. The model is **tested** using the test dataset.  This results in a **prediction** that includes a **score** for each hotel instance.  The score is used to determine the ranking relative to other hotels within the same query (e.g. group).  The predictions are then **evaluated** by examining metrics; specifically the [Normalized Discounted Cumulative Gain](https://en.wikipedia.org/wiki/Discounted_cumulative_gain) (NDCG).
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+5. The final step is to **consume** the model to perform ranking predictions for new incoming hotel searches.
+
+### 1. Label Data
+To label the data with ideal ranking values, the sample follows [Expedia's evaluation guidance](https://www.kaggle.com/c/expedia-personalized-sort/overview/evaluation):
+
+* 0 - The user neither clicked on this hotel nor purchased\booked a room at this hotel.
+* 1 - The user clicked through to see more information on this hotel.
+* 2 - The user purchased\booked a room at this hotel.
+
+Expedia's dataset includes both **Click_Bool** and **Booking_Bool** columns that indicate whether the user has clicked or purchased/booked a hotel.  Applying the above guidelines to these columns, we create a new **Label** column that contains values {0, 1, 2} for each hotel search result.
```
justinormont: May want to mention the the relevance gains {0, 1, 5} that these {0, 1, 2} relevance scores map to.

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+### 2. Split Data
+ With the data properly labeled, it is ready to be split into the train/test datasets.  When splitting the data, it's important to make sure that the hotel search results for a given query aren't split across the two datasets.  This would cause label leakage where the same query in our training dataset also exists within the testing dataset.
+
+ Refer to the following code which shows how to split the data:
+
+ ```CSharp
+// When splitting the data, 20% is held for the test dataset.
+// To avoid label leakage, the GroupId (e.g. search\query id) is specified as the samplingKeyColumnName.  
+// This ensures that if two or more hotel instances share the same GroupId, that they are guaranteed to appear in the same subset of data (train or test).
+TrainTestData trainTestData = mlContext.Data.TrainTestSplit(labeledData, testFraction: .2, samplingKeyColumnName: nameof(HotelData.GroupId), seed: 1);
```
justinormont: Leading zero:
```suggestion
TrainTestData trainTestData = mlContext.Data.TrainTestSplit(labeledData, testFraction: 0.2, samplingKeyColumnName: nameof(HotelData.GroupId), seed: 1);
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+        {
+            output.Label = 0;
+        }
+    };
+
+    return mapper;
+}
+`````
+### 2. Split Data
+ With the data properly labeled, it is ready to be split into the train/test datasets.  When splitting the data, it's important to make sure that the hotel search results for a given query aren't split across the two datasets.  This would cause label leakage where the same query in our training dataset also exists within the testing dataset.
```
justinormont: Word smithing...
```suggestion
 With the data properly labeled, it is ready to be split into the train/test datasets.  When splitting the dataset, it's 
 important to make sure all of the results for a single hotel search remain in the same dataset split. This would cause data leakage where the same query in our training dataset also exists within the testing dataset.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
justinormont: Leading zero...
```suggestion
TrainTestData trainTestData = mlContext.Data.TrainTestSplit(labeledData, testFraction: 0.2, samplingKeyColumnName: nameof(HotelData.GroupId), seed: 1);
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+ Refer to the following code which shows how to split the data:
+
+ ```CSharp
+// When splitting the data, 20% is held for the test dataset.
+// To avoid label leakage, the GroupId (e.g. search\query id) is specified as the samplingKeyColumnName.
```
justinormont:
```suggestion
// To avoid data leakage, the GroupId (e.g. search\query id) is specified as the samplingKeyColumnName.  
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+// Save the train dataset to a file to make it faster to load in subsequent runs.
+using (var fileStream = File.Create(testDatasetPath))
+{
+    mlContext.Data.SaveAsText(testData, fileStream, separatorChar: ',', headerRow: true, schema: true);
+}
+`````
+
+### 3. Train Model
+This sample trains the model using the LightGbmRankingTrainer which relies on the LightGbm Lambdarank algorithm.  The model requires the following input columns:
```
justinormont:
```suggestion
This sample trains the model using the LightGbmRankingTrainer which relies on the LightGBM LambdaRank algorithm.  The model requires the following input columns:
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+* Group Id - Column that contains the group id for each data instance.  Data instances are contained in logical groupings and each group has an identifier known as the group id.  In the case of the Expedia dataset, hotel search results are grouped by their corresponding query where the group id corresponds to the query or search id.  The input group id data type must be [key type](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.keydataviewtype).
```
justinormont:
```suggestion
* Group Id - Column that contains the group id for each data instance.  Data instances are contained in logical groupings representing all candidate results in a single query and each group has an identifier known as the group id.  In the case of the Expedia dataset, hotel search results are grouped by their corresponding query where the group id corresponds to the query or search id.  The input group id data type must be [key type](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.keydataviewtype). 
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+{
+    mlContext.Data.SaveAsText(testData, fileStream, separatorChar: ',', headerRow: true, schema: true);
+}
+`````
+
+### 3. Train Model
+This sample trains the model using the LightGbmRankingTrainer which relies on the LightGbm Lambdarank algorithm.  The model requires the following input columns:
+
+* Group Id - Column that contains the group id for each data instance.  Data instances are contained in logical groupings and each group has an identifier known as the group id.  In the case of the Expedia dataset, hotel search results are grouped by their corresponding query where the group id corresponds to the query or search id.  The input group id data type must be [key type](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.keydataviewtype). 
+* Label - Column that contains the deal rank (e.g. degree of relevance) of each data instance where higher values indicate higher relevance.  The input label data type must be [key type](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.keydataviewtype) or [Single](https://docs.microsoft.com/en-us/dotnet/api/system.single).
```
justinormont:
```suggestion
* Label - Column that contains the relevance label of each data instance where higher values indicate higher relevance.  The input label data type must be [key type](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.keydataviewtype) or [Single](https://docs.microsoft.com/en-us/dotnet/api/system.single). 
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+* Group Id - Column that contains the group id for each data instance.  Data instances are contained in logical groupings and each group has an identifier known as the group id.  In the case of the Expedia dataset, hotel search results are grouped by their corresponding query where the group id corresponds to the query or search id.  The input group id data type must be [key type](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.keydataviewtype). 
+* Label - Column that contains the deal rank (e.g. degree of relevance) of each data instance where higher values indicate higher relevance.  The input label data type must be [key type](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.keydataviewtype) or [Single](https://docs.microsoft.com/en-us/dotnet/api/system.single). 
+* Features - The columns that are influential in determining the relevance\rank of a data instance.  The input feature data must be a fixed size vector of type [Single](https://docs.microsoft.com/en-us/dotnet/api/system.single).
```
justinormont:
```suggestion
* Features - The columns that are influential in determining the relevance\rank of a data instance. The input feature data must be a fixed size vector of type [Single](https://docs.microsoft.com/en-us/dotnet/api/system.single).
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+When the trainer is set, **custom gains** are used to apply weights to each of the labeled rank values.  As described earlier in the sample, the potential label rank values are {0, 1, 2} which directly correlates to the specified gains {0, 1, 5}.  This helps to ensure that the model places more emphasis on ranking hotel search results labeled with 2 (e.g. signifies the user purchased\booked the hotel) so that they are positioned higher when compared to results labeled with 0 or 1.
```
justinormont: Extra spaces:
```suggestion
When the trainer is set, **custom gains** are used to apply weights to each of the labeled rank values. As described earlier in the sample, the potential label rank values are {0, 1, 2} which directly correlates to the specified gains {0, 1, 5}. This helps to ensure that the model places more emphasis on ranking hotel search results labeled with 2 (e.g. signifies the user purchased\booked the hotel) so that they are positioned higher when compared to results labeled with 0 or 1.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+    .Select(s => s.Name)
+    .Where(c =>
+        c == nameof(HotelData.Price_USD) ||
+        c == nameof(HotelData.Promotion_Flag) ||
+        c == nameof(HotelData.Prop_Id) ||
+        c == nameof(HotelData.Prop_Brand) ||
+        c == nameof(HotelData.Prop_Review_Score))
+    .ToArray();
+
+// Set trainer options.
```
justinormont: Recommend doing a mass removal of trailing spaces:
```suggestion
// Set trainer options.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+// Create an Estimator and transform the data:
+// 1. Concatenate the feature columns into a single Features vector.
+// 2. Create a key type for the label input data by using the value to key transform.
+// 3. Create a key type for the group input data by using a hash transform.
+IEstimator<ITransformer> dataPipeline = mlContext.Transforms.Concatenate(FeaturesVectorName, featureCols)
+    .Append(mlContext.Transforms.Conversion.MapValueToKey(nameof(HotelData.Label)))
+    .Append(mlContext.Transforms.Conversion.Hash(nameof(HotelData.GroupId), nameof(HotelData.GroupId), numberOfBits: 20));
+
+// Set the LightGbm Lambdarank trainer.
```
justinormont:
```suggestion
// Set the LightGBM LambdaRank trainer.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+`Evaluate()` compares the predicted values for the test dataset and produces various metrics, such as accuracy, you can explore.  Specifically, we can gauge the accuracy of our model using Discounted Cumulative Gain (DCG) and Normalized Discounted Cumulative Gain (NDCG) which are included in the `RankingMetrics` returned by `Evaluate()`. 
+
+When evaluating the `RankingMetrics` for this sample's model, you'll notice that the following metrics are reported for DCG and NDCG (the values that you see when running the sample will be similar to these):
+* DCG - @1:1.0191, @2:1.5128, @3:1.8371, @4:2.0922, @5:2.2982, @6:2.4641, @7:2.6051, @8:2.7240, @9:2.8234, @10:2.9133
+
+* NDCG - @1:0.1184, @2:0.1719, @3:0.2082, @4:0.2372, @5:0.2608, @6:0.2798, @7:0.2960, @8:0.3096, @9:0.3210, @10:0.3314
+
+The NDCG values are most useful to examine since this allows us to compare accuracy across different queries.  The potential value of NDCG ranges from **0.0** to **1.0**, with 1.0 being a perfect model that exactly matches the ideal ranking.
```
justinormont:
```suggestion
The NDCG values are most useful to examine since this allows us to compare accuracy across different datasets.  The potential value of NDCG ranges from **0.0** to **1.0**, with 1.0 being a perfect model that exactly matches the ideal ranking.  
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+* DCG - @1:1.0191, @2:1.5128, @3:1.8371, @4:2.0922, @5:2.2982, @6:2.4641, @7:2.6051, @8:2.7240, @9:2.8234, @10:2.9133
+
+* NDCG - @1:0.1184, @2:0.1719, @3:0.2082, @4:0.2372, @5:0.2608, @6:0.2798, @7:0.2960, @8:0.3096, @9:0.3210, @10:0.3314
+
+The NDCG values are most useful to examine since this allows us to compare accuracy across different queries.  The potential value of NDCG ranges from **0.0** to **1.0**, with 1.0 being a perfect model that exactly matches the ideal ranking.  
+
+With this in mind, let's look at our model's values for NDCG.  In particular, let's look at the value for **NDCG@10** which is **.3314**. This is the average NDCG for a query returning the top **10** hotel search results.  While **.3314** may seem low compared to **1.0**, a more realistic goal is to reach **.5407** which is the score of the first place winner in [Expedia's Personalize Hotel Search contest on Kaggle](https://www.kaggle.com/c/expedia-personalized-sort/leaderboard).  To increase the model's accuracy, we would need to experiment with feature engineering to continue to improve our model.
```
justinormont: Leading zeros:
```suggestion
With this in mind, let's look at our model's values for NDCG.  In particular, let's look at the value for **NDCG@10** which is **0.3314**. This is the average NDCG for a query returning the top **10** hotel search results.  While **0.3314** may seem low compared to **1.0**, a more realistic goal is to reach **0.5407** which is the score of the first place winner in [Expedia's Personalize Hotel Search contest on Kaggle](https://www.kaggle.com/c/expedia-personalized-sort/leaderboard).  To increase the model's accuracy, we would need to experiment with feature engineering and model hyperparameters to continue to improve our model.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+IDataView testData = mlContext.Data.LoadFromTextFile<HotelData>(testDatasetPath, separatorChar: ',', hasHeader: true);
+IDataView predictions = model.Transform(testData);
+
+[...]
+
+// Evaluate the metrics for the data using NDCG; by default, metrics for the up to 3 search results in the query are reported (e.g. NDCG@3).
+RankingMetrics metrics = mlContext.Ranking.Evaluate(scoredData);
+`````
+
+### 5.  Consume Model
```
justinormont:
```suggestion
### 5. Consume Model
```

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+// Predict rankings.
+IDataView predictions = predictionPipeline.Transform(exampleData);
+
+// In the predictions, get the scores of the hotel search results included in the first query (e.g. group).
+IEnumerable<HotelPrediction> hotelQueries = mlContext.Data.CreateEnumerable<HotelPrediction>(predictions, reuseRowObject: false);
+var firstGroupId = hotelQueries.First<HotelPrediction>().GroupId;
+IEnumerable<HotelPrediction> firstGroupPredictions = hotelQueries.Take(50).Where(p => p.GroupId == firstGroupId).OrderByDescending(p => p.PredictedRank).ToList();
+
+// The individual scores themselves are NOT a useful measure of accuracy; instead, they are used to determine the ranking where a higher score indicates a higher ranking.
```
justinormont:
```suggestion
// The individual scores themselves are NOT a useful measure of result quality; they are only useful as a relative measure to other scores in the group. The scores are used to determine the ranking where a higher score indicates a higher ranking vs another candidate result.
```

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Common/ConsoleHelper.cs
nicolehaugen: I don't know for sure - I actually had tried setting this value to greater than 10 (specifically 38 since this was what was used for the Kaggle contest).   It turns out that the code has explicit checks for greater than 10 and throws an exception (which is why I'm also doing this check).  And, even more confusing is that there are comments in the ML .NET code indicating that a value up to 100 should be allowed (but with a ? mark behind it).  I talked with @eerhardt about this and he suspects that there must be a reason that they've limited it to 10.  I plan to log an issue for this.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
nicolehaugen: Yes you do need a Kaggle account for this - I had assumed that all Kaggle datasets required you have an account.  But, based on your comment, that must not be the case.  You probably noticed but I have sent an email to CELA to learn if there is a possibility of using this dataset.  If it turns out I can't, then I will look at using the benchmark dataset.  Overall, I feel that the hotel data is a very compelling scenario which is why I decided to try pursuing it.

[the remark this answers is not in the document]
nicolehaugen: See comment above - I will use this as a substitution if I can't get approval from CELA on the expedia dataset.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
nicolehaugen: I also replaced other references to "accuracy" with "quality".

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
nicolehaugen: I replaced all references to "ideal ranking value" to "relevance score".

For your point on "This helps to avoid converting a ranked list directly into equally sized bins of {0, 1, 2, 3, 4}" - can you explain a bit more why this needs to be avoided?  I'd like to understand this point and then also explain it in the sample.  For example, would the model potentially end up predicting ranks that are nearly the same for each result?

Good point about reusing scores - I inadvertently left this out and will add it.

nicolehaugen: I haven't tried it (I lifted this point from the TLC documentation and assumed it was still applicable).  I will verify it to confirm.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Common/ConsoleHelper.cs
justinormont: I don't expect there should be a reason to limit the number of slots used in the DCG/NDCG calculation.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
justinormont: I'd recommend all demos have datasets that can be downloaded automatically (no user action needed). Will CELA let us host the dataset in our CDN?

Granted Kaggle is pretty awesome and folks should hone there data science skill there. I think forcing our users to sign up for an external account before trying demo code is a bit rough.

We will have more ranking samples of course. For instance a short sample that teaches less aspects but is easy to modify for towards their own datasets. And perhaps a sample which does string comparisons (query vs. documents) for feature engineering.

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
nicolehaugen: Note that the actual documentation provided for this parameter says the following: "This can be used to ensure no label leakage from the train to the test set."

This obviously is a minor issue in the doc, but sounds like the way it's currently written is incorrect?

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
nicolehaugen: I followed up on this by changing my label values to strings ("Perfect", "Fair", "Bad") - when I attempt to train the model I get an Invalid Op exception saying the following, so it looks like this isn't supported.  Does a bug need to be logged for this?

"Splitter/consolidator worker encountered exception while consuming source data.
"Could not parse value Bad in line 62, column Label".

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
justinormont: I'm unsure if it's considered a bug or design choice at this point. I find `{"Perfect", "Excellent", "Good", "Fair", or "Bad"}` to be more user friendly. But the design of the API isn't my area.

For your exact error, it may be that you have to change your Label column from `Int` to `String`.

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
justinormont: It is a form of leakage --  I think I'd group the various styles of leakage in the following way...

Leakage styles:
* **Feature availability** (columns which aren't available at prediction time -- e.g. a MonthySalary columns when predicting YearlySalary; or MinutesLate when predicting IsLate; or more subtly NumOfLatePayments when predicting ShouldGiveLoan)
* Non-iid data:
  * **Data leakage** (e.g. splitting a time-series dataset randomly instead of newer data in test set; or MinMax normalizing a dataset then splitting)
  * **Duplicate rows** between train/validation/test (e.g. oversampling a dataset to pad its size before splitting; e.g. different rotations of an single image; bootstrap sampling before splitting; or duplicating rows to up sample minority class) 
  * **Group leakage** -- not including a grouping split column (e.g. Andrew Ng's team had 100k x-rays of 30k patients, meaning ~3 images per patient. They did random splitting instead of ensuring that all images of a patient was in the same split. Hence the model partially memorized the patients instead of learning to recognize pneumonia in chest x-rays. Revised paper had a noticeable drop in scores.)

The first level is if it's column based, or row based leakage. These leakages cause you to misestimate how well your model will perform in production.

This style of leakage is a style of row based leakage, where the lack of a `samplingKeyColumnName` column causes group leakage which may also show up as data leakage or duplicate rows. I'm uncertain which I'd call it; or if my ontology of leakage styles is sufficient.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
```
eerhardt: All the elements in this ItemGroup shouldn't be necessary. This whole group can be deleted.

eerhardt: oh, I see you have a `.csproj` nested under another `.csproj`. This isn't typical. Can they be in sibling folders instead?

---
In reply to: [298226619](https://github.com/dotnet/machinelearning-samples/pull/533#discussion_r298226619) [](ancestors = 298226619)

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,218 @@
...
+The following code is used to train the model:
+
+```CSharp
+const string FeaturesVectorName = "Features";
+
+// Load the training dataset.
+IDataView trainData = mlContext.Data.LoadFromTextFile<HotelData>(trainDatasetPath, separatorChar: ',', hasHeader: true);
+
+// Specify the columns to include in the feature input data.
+var featureCols = trainData.Schema.AsQueryable()
```
eerhardt: Could this simply be:

```
var featureCols = new[] { nameof(HotelData.Price_USD), nameof(HotelData.Promotion_Flag), nameof(HotelData.Prop_Id), nameof(HotelData.Prop_Brand), nameof(HotelData.Prop_Review_Score)};
```

?

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
```diff
@@ -0,0 +1,190 @@
...
+            }
+
+            Console.ReadLine();
+        }
+
+        static void PrepDatasets(MLContext mlContext, string assetPath, string originalDatasetPath, string trainDatasetPath, string testDatasetPath)
+        {
+            const string DatasetUrl = "https://www.kaggle.com/c/expedia-personalized-sort/download/data.zip";
+
+                if (!File.Exists(trainDatasetPath) || !File.Exists(testDatasetPath))
```
eerhardt: (nit) the whitespace is off here.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
```diff
@@ -0,0 +1,190 @@
...
+                {
+                    if (!File.Exists(originalDatasetPath))
+                    {
+                        throw new InvalidOperationException($"This samples requires the Expedia dataset.  Please ensure that you have downloaded and extracted the contents of the .zip file to the following directory: {assetPath}. The .zip file can be downloaded from here: {DatasetUrl}");
+                    }
+
+                    Console.WriteLine("===== Prepare the testing/training datasets =====");
+
+                    // Load dataset using TextLoader by specifying the type name that holds the data's schema to be mapped with datasets.
+                    IDataView data = mlContext.Data.LoadFromTextFile<HotelData>(originalDatasetPath, separatorChar: ',', hasHeader: true);
```
eerhardt: Where do `HotelData` and `HotelPrediction` come from? I don't see them in this PR.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
nicolehaugen: I have now switched the dataset to use the one you recommended for ranking search engine results.

#resolved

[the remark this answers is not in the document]
nicolehaugen: Same comment as above - am now using this dataset. #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
nicolehaugen: #resolved

nicolehaugen: Based on my above comment, I have removed the mention of using strings as label values. #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
nicolehaugen: #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking.csproj
nicolehaugen: Yes, that was my mistake when I was copying and pasting some things around while cleaning up my solution.  I have deleted the extra .csproj file.

nicolehaugen: #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
nicolehaugen: I have since moved to a new dataset where there are 136 columns - as a result, this code is now: var featureCols = trainData.Schema.AsQueryable()
    .Select(s => s.Name)
    .Where(c =>
        c != nameof(SearchResultData.Label) &&
        c != nameof(SearchResultData.GroupId))
    .ToArray();

#resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
nicolehaugen: #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
nicolehaugen: They are in the DataStructures directory - I see them in the PR...maybe you overlooked it?  Anyway, since moving to the new dataset, these files have been removed and replaced with new types.

nicolehaugen: #resolve

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/PersonalizedRanking.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Remove="DataStructures\HotelData.cs" />
```
eerhardt: These can all be removed, can't they? I don't see a file `Program_old.cs` or `Mapper.cs`

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Common/ConsoleHelper.cs
```diff
@@ -0,0 +1,62 @@
+﻿using Microsoft.ML;
+using Microsoft.ML.Data;
+using PersonalizedRanking.DataStructures;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace PersonalizedRanking.Common
```
justinormont: Renaming:
I might recommend a title for this sample: "MSLR-WEB10K Ranking" or more simply, "Web Ranking".

I don't think there is [personalization](https://en.wikipedia.org/wiki/Personalized_search) in this dataset. Personalization in the ranking area generally means the each user gets individualized search results. In this dataset, there is not information about the user. This information would generally include { topics of interest of the user, demographics of the user, current location of the user, etc }.

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Linq;
+using System.Net;
+
+namespace PersonalizedRanking
+{
+    class Program
+    {
+        const string AssetsPath = @"../../../Assets";
+        const string TrainDatasetUrl = "https://aka.ms/mlnet-resources/benchmarks/MSLRWeb10KTrain720kRows.tsv";
+        const string TestDatasetUrl = "https://aka.ms/mlnet-resources/benchmarks/MSLRWeb10KTest240kRows.tsv";
```
justinormont: Doesn't hurt to list the validation dataset. This would let users experiment on the dataset by iterating using the scores on the validation set, then testing their final solution, exactly once, on the test dataset.

Generally the DS pattern for ML should look like:
1. Train on training set, get metrics from validation set
2. (iterate many times until happy with pipeline -- goto 1)
3. Train on found pipeline on combined train+validate set, get metrics on test set (exactly once) -- this is your final metrics
4. Retrain  pipeline on all data train+validate+test set. Send this newly created model to production.

The final estimate of how well your model will do in production is the metrics from step (3). The final model for production, trained on all available data, is trained at step (4).

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
```diff
@@ -0,0 +1,142 @@
...
+* Travel Agency - Provide a list of hotels with those that are most likely to be purchased\booked by the user positioned highest in the list.
+* Shopping - Display items from a product catalog in an order that aligns with a user's shopping preferences.
+* Recruiting - Retrieve job applications ranked according to the candidates that are most qualified for a new job opening.
+
+Ranking is useful to any scenario where it is important to list items in an order that increases the likelihood of a click, purchase, reservation, etc.
+ 
+In this sample, we show how to apply ranking to search engine results.  To perform ranking, there are two algorithms currently available - FastTree Boosting (FastRank) and Light Gradient Boosting Machine (LightGBM).  We use the LightGBM's LambdaRank implementation in this sample to automatically build an ML model to predict ranking. 
+
+## Dataset
+The training and testing data used by this sample is based on a public [dataset provided by Microsoft](https://www.microsoft.com/en-us/research/project/mslr/) originally provided Microsoft Bing.
```
justinormont: Added note about the licensing of the dataset:
```suggestion
The training and testing data used by this sample is based on a public [dataset provided by Microsoft](https://www.microsoft.com/en-us/research/project/mslr/) originally provided Microsoft Bing. The dataset is released under a [CC-by 4.0](https://creativecommons.org/licenses/by/4.0/) license.

This information is also available in the datasets readme in the main repo -- https://github.com/dotnet/machinelearning/blob/056c60479304a3b5dbdf129c9bc6e853322bb090/test/data/README.md#mslr-web10k-mslr-web30k

May want to take the citation from there also.

File: samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/DataStructures/SearchResultPrediction.cs [hunk 35]
```diff
@@ -0,0 +1,17 @@
...
+namespace PersonalizedRanking.DataStructures
+{
+    // Representation of the prediction made by the model (e.g. ranker).
+    public class SearchResultPrediction
+    {
+        public uint GroupId { get; set; }
+
+        public uint Label { get; set; }
+
+        // Prediction made by the model that is used to indicate the relative ranking of the benchmark data instances.
```
justinormont:
```suggestion
        // Prediction made by the model that is used to indicate the relative ranking candidate search results
```

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Common/ConsoleHelper.cs
nicolehaugen: #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/PersonalizedRanking.csproj
nicolehaugen: yep - i missed that - thanks! #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Common/ConsoleHelper.cs
nicolehaugen: #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/README.md
nicolehaugen: #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/DataStructures/SearchResultPrediction.cs
nicolehaugen: #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/PersonalizedRanking.csproj
nicolehaugen: #resolved

samples/csharp/getting-started/Ranking_PersonalizedSort/PersonalizedRanking/Program.cs
nicolehaugen: #Resolved

PR #562, opened 2019-07-18
Fix security vulnerability

bamurtaugh:
-Updated lodashtemp to version 4.5.0


bamurtaugh:
Updated lodash.template by hand since GitHub couldn't automatically create a fix/PR. When I ran the web app, it behaves well. Only changed "lodash.template" and not "lodash.templatesettings" since the latter seems to be a different dependency that doesn't need to be updated.

PR #587, opened 2019-07-25
Updated sample to maintain consistency with getting started app

luisquintanilla:
List of changes

- Renamed `PaintImages` method
- Removed `CreateEmptyDataView` method and replaced with one-liner.
- Reanemd `YoloWinMlParser` to `YoloOutputParser` and replaced existing code with getting-started ONNX object detection sample code. 
- Added required domain classes `DimensionsBase`, `CellDimensions` and updated `YoloBoundingBox` class definition.


bamurtaugh:
Looks good when I run it. 

However, last week we completed migration to ML.NET version 1.2.0 and preview version 0.14.0. I noticed some older/hard-coded versions in this project (i.e. OnnxObjectDetectionE2EApp.csproj has hardcoded "Version=0.13.0" and "Version="0.12.0," Directory.Build.props hasn't been updated). 

Could you perhaps pull the versioning changes from master where we updated to v1.2, or manually update your versions of ML.NET and ML.NET Preview in the .csproj and the Directory.build.props? Just want to make sure everything works correctly in the newest versions!

luisquintanilla:
@bamurtaugh thanks for reviewing. Done. I also updated the csproj of the getting-started ONNX sample as well. Let me know if there's anything else.

samples/csharp/end-to-end-apps/DeepLearning_ObjectDetection_Onnx/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+
+  <PropertyGroup>
+    <MicrosoftMLVersion>1.2.0</MicrosoftMLVersion>
```
bamurtaugh: Is there a specific reason we are introducing an additional Directory.Build.props file here? It has the same updated version information found in the [Directory.Build.props in the broader samples folder](https://github.com/dotnet/machinelearning-samples/blob/master/samples/Directory.Build.props).

CESARDELATORRE: Agree. If the version is the same than the rest of the samples it should use the file from the outer folder. A specific  Directory.Build.props file for a sample should only be used if the version being used is different than the version of the rest of the samples.

luisquintanilla: Just deleted the file and tested. Versions resolved to those in top level folder.

PR #589, opened 2019-07-29
RandomizedPCA Anomaly Detection fraud detection sample

colbylwilliams:
- RandomizedPCA sample based on [this Binary Classification sample](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/BinaryClassification_CreditCardFraudDetection)
- Note that final completion of this sample is waiting on fix of [this issue](https://github.com/dotnet/machinelearning/issues/3990)


bamurtaugh:
Looks good when I run it.

However, a couple weeks ago, we completed migration to ML.NET version 1.2.0 and preview version 0.14.0. I noticed the Directory.Build.props file in this project is using MicrosoftMLVersion 1.0.0 and MicrosoftMLPreviewVersion 0.12.0. Can you please verify that everything works with the most updated versions as well?

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/CreditCardFraudDetection.Predictor/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+            modelPredictor.RunMultiplePredictions(numberOfPredictions: 5);
+
+            Console.WriteLine("=============== Press any key ===============");
+            Console.ReadKey();
+        }
+
+
+        public static void CopyModelAndDatasetFromTrainingProject(string trainOutput, string assetsPath)
+        {
+            if (!File.Exists(Path.Combine(trainOutput, "testData.csv")) ||
```
bamurtaugh: If testData.csv doesn't exist, wouldn't that be a different error than not running the training project first? Would you want to give a different/separate error for that case?

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/CreditCardFraudDetection.Predictor/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+                !File.Exists(Path.Combine(trainOutput, "randomizedPca.zip")))
+            {
+                Console.WriteLine("***** YOU NEED TO RUN THE TRAINING PROJECT FIRST *****");
```
bamurtaugh: Is there a reason we don't include a copy of randomizedPca.zip for the user already (i.e. is it too large of a file to upload to GitHub)? For the other samples, I believe we include a copy of the model/files produced from training, and the user doesn't have to train first themselves to get the predictor to work.

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/Readme.md
```diff
@@ -0,0 +1,194 @@
+# Fraud detection in credit cards based on anomaly (or outlier) detection and existing PCA-transformed dataset
+
+| ML.NET version | API type    | Status     | App Type         | Data type | Scenario        | ML Task           | Algorithms     |
+|----------------|-------------|------------|------------------|-----------|-----------------|-------------------|----------------|
+| v1.1.0         | Dynamic API | Up-to-date | Two console apps | .csv file | Fraud Detection | Anomaly Detection | Randomized PCA |
```
bamurtaugh: This would now be v1.2.0 I believe? 😄

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/Readme.md
```diff
@@ -0,0 +1,194 @@
...
+To solve this problem, first you need to build a machine learning model. Then you train the model on existing training data, evaluate how good its accuracy is, and lastly you consume the model (deploying the built model in a different app) to predict a fraud for a sample credit card transaction.
+
+![Build -> Train -> Evaluate -> Consume](../shared_content/modelpipeline.png)
+
+
+### 1. Build model
+
+Building a model includes:
+
+- Preapre the data and split data for training and tests
```
bamurtaugh:
```suggestion
- Prepare the data and split data for training and tests
```

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/CreditCardFraudDetection.Predictor/Program.cs
colbylwilliams: As it stands now, the training project splits the original dataset into train/test.  So both the trained model (`reandomizedPca.zip`) and the test data csv (`testData.csv`) are copied into the `CreditCardFraudDetection.Trainer`s output directory upon running the trainer project.  So as it currently works, it is the same error, as `CreditCardFraudDetection.Predictor` is looking for both files output from the trainer project.

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/CreditCardFraudDetection.Predictor/Program.cs
colbylwilliams: I based this on how it was done in the [BinaryClassification version of this sample](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/BinaryClassification_CreditCardFraudDetection).  Although, it actually looks like that sample does [commit the model](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/BinaryClassification_CreditCardFraudDetection/CreditCardFraudDetection.Predictor/assets/input) into the input directory of the sample's Predictor project.  However, as mentioned above, both the test data csv and the model are generated by the Trainer project (in both samples) and looks like the BinaryClassification version of this sample doesn't commit that csv either.  So even in the BinaryClassification sample, the Trainer project will need to be run before the Predictor sample will work.

I'm happy to follow you guidance here.  If we want to commit the model and test data set into the Predictor project, I can definitely make the change, but we should probably do it for both samples.  I believe the longer-term goal is to have all these samples run as a single project vs. separate Trainer/Predictor projects.  But again, that's a change that should be made in both the CreditCardFraudDetection samples, and may be outside the scope of this PR.

Let me know how you want to move forward.

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/Readme.md
colbylwilliams: resolved in fcf77240b0acf610fe81cb01a68344d2d2f84d07

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/CreditCardFraudDetection.Predictor/Program.cs
bamurtaugh: Makes sense!

samples/csharp/getting-started/AnomalyDetection_CreditCardFraudDetection/CreditCardFraudDetection.Predictor/Program.cs
bamurtaugh: @CESARDELATORRE Would love to get your thoughts here!

CESARDELATORRE: This concrete sample has two projects. The predictor/scoring project can have the model.ZIP file. But the training project doesn't need to have it and since the code is split in two projects it is even better that the training project doesn't have it.
We can merge it as it is now. :)

CESARDELATORRE: Btw, never mind. For this case it might be better if the scoring/client app copies the model so it takes the latest training. We'll change the code so it is only copying the .zip model and the test dataset. The scoring project doesn't need the training dataset and git ignore files that are being copied currently. But this is a minor improvement for clarity. 

Thanks! 👍

PR #591, opened 2019-07-30
Adding assets for ONNX object detection tutorial

luisquintanilla:
Added zipped "assets" directory to make it easier for users to download the files for use in the tutorial.


DevLob-zz:
is ONNX  is valid for page layout segmentation 



if i have the original image and label regions with different colors could it work if i give it an image and predict the region from it

CESARDELATORRE:
Not sure what you mean. ONNX is simply an standard format for ML models.
What you can do really depends on the ML model you have which in this case dealing with images could have been created with TensorFlow, PyTorch or any other DeepLearning framework.

PR #601, opened 2019-08-08
Adding Azure Functions Sentiment Analysis E2E App Sample

luisquintanilla:
Adding Azure Functions Sentiment Analysis E2E App based on How-To Guide 

https://docs.microsoft.com/en-us/dotnet/machine-learning/how-to-guides/serve-model-serverless-azure-functions-ml-net

Updates:

- Uses `FunctionsStartup` and `IFunctionsHostBuilder`.

To-Do:

- [x] Create README


CESARDELATORRE:
For the "cross-cutting" samples, such as Web apps, Database, etc. we're trying not to lead with the ML Task but with the cross-cutting feature, such as:
- DatabaseIntegration 
- ScalableMLModelOnWebAPI 
- etc.


Could you change the Root folder of the sample to just something like:

"**ScalableMLModelOnAzureFunction**"

PR #609, opened 2019-08-16
WIP - Model Builder Sentiment Analysis Razor Pages Sample

luisquintanilla:
To-Do:

- [ ] Create README


samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazorML.ConsoleApp/ModelBuilder.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Linq;
+using Microsoft.ML;
+using Microsoft.ML.Data;
+using SentimentRazorML.Model.DataModels;
+
+namespace SentimentRazorML.ConsoleApp
+{
+    public static class ModelBuilder
+    {
+        private static string TRAIN_DATA_FILEPATH = @"C:\Users\luquinta.REDMOND\Downloads\wikipedia-detox-250-line-data.tsv";
```
justinormont: Local path should be relative to the project.

samples/modelbuilder/BinaryClassification_Sentiment_Razor/README.md
```diff
@@ -0,0 +1,73 @@
...
+# Sentiment Analysis: Razor Pages sample optimized for scalability and performance when running/scoring an ML.NET model built with Model Builder (Using the new '.NET Core Integration Package for ML.NET')
+
+
+| ML.NET version | Status                        | App Type    | Data type | Scenario            | ML Task                   | Algorithms                  |
+|----------------|-------------------------------|-------------|-----------|---------------------|---------------------------|-----------------------------|
+| v1.3.1           | Up-to-date | Razor Pages | Single data sample | Sentiment Analysis | Binary classification | Linear Classification |
+
+# Goal
+
+The goal is to be able to make **SENTIMENT ANALYSIS** prediction/detection of what the user is writing in a very **UI interactive app** in the client side and running an **ML.NET model**(Sentiment analysis based on binary-classification) built with Model Builder in the server side.
```
justinormont: Missing space:
```suggestion
The goal is to be able to make **SENTIMENT ANALYSIS** prediction/detection of what the user is writing in a very **UI interactive app** in the client side and running an **ML.NET model** (Sentiment analysis based on binary-classification) built with Model Builder in the server side.
```

samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazorML.ConsoleApp/ModelBuilder.cs
justinormont: Also this dataset is way too small to make a useful sentiment model. Might recommend a larger sentiment dataset. 

As a start, we have the full size (~70MB) Wikipedia Detox dataset. 

Though the Wiki Detox datasets is a rather specific version of "sentiment" of "is _personal attack_ in the context of Wikipedia editor's comments" (subject of the attack must be another Wikipedia editor and directed at them). 

Made up examples:
* "You suck" == true _-- not "you" btw, you're awesome!_
* "Jeffrey Dahmer was a terrible person" == false (3rd person)
* "2019-08-14 was the 4th largest drop in the stock market" == false (not attacking a person).

justinormont: We can pull the full sized Wikipedia Detox dataset from our CDN: 
http://aka.ms/tlc-resources/benchmarks/WikiDetoxAnnotated160kRows.tsv

CESARDELATORRE: We're using medium size WikiDetox datasets in other samples. You can use those instead of the 250 rows dataset which is too small..

luisquintanilla: Hmm. good catch @justinormont . This was part of the auto-generated project. 

Question:

Should the auto-generated project hard-code the path where the data lives or should it copy the data to the auto-generated project, include it in the output directory and reference it from within the project? This way, if the user deletes the original data file, the project still works.

luisquintanilla: Thanks for the feedback. I agree the current dataset is too small. Few concerns with this dataset. 

1) Currently in model builder as far as I'm aware there is no way to exclude columns so the user would have to do pre-processing on the data to take out the columns themselves. 
2) Since it's part of the CDN, is it safe to assume this link won't change and will be up for some time?

justinormont: Yes, the CDN is meant to continue to exist. It serves up various parts of ML.NET, like the word embedding files, and various datasets for our benchmarks. The aka.ms link exists so in the future we can re-locate the files by changing where the link points to.

Model Builder will have a method of dropping columns in the future (it may already, but I haven't checked). 

It would be nice to drop `sample` & `split` as they could (but not necessarily do) leak information as the features wouldn't be available when running the model. The other columns are dropped in the 250-line file just to make the example more terse.

Side note on the 250-line WikiDetox dataset, it has duplicate rows in the testset which are in the training set. Specifically, these rows in the testset are also in the train:
```
1          Stop trolling, zapatancas, calling me a liar merely demonstartes that you arer Zapatancas. You may choose to chase every legitimate editor from this site and ignore me but I am an editor with a record that isnt 99% trolling and therefore my wishes are not to be completely ignored by a sockpuppet like yourself. The consensus is overwhelmingly against you and your trollin g lover Zapatancas,
1         == Conflict of interest ==  You are a person who is doing some sort of harm to this lady Saman Hasnain.. It is apparent that you are making sure that her name is defamed.... Okay no problem... Will get a better source... you are playing dirty... DOG Sonisona
1         ==RUDE== Dude, you are rude upload that carl picture back, or else.
1         ==You're cool==  You seem like a really cool guy... *bursts out laughing at sarcasm*.
1        :the category was unnecesary, as explained in my edit summary. Your threats are disgrace to wikipedia.
1       " :You call MacDonald's a part of your ""culture""? Nonsense! Spend some 10 years in France, and then you will have a hint of what Culture is!   "
1       Grow up you biased child.
```
7 of 18 rows are duplicates. Also, the testset is too small to have any utility. We should remove these duplicate rows, and also grow this dataset to a useful size, or discontinue it's use including in unit tests. Users have sent feedback that their created models aren't useful, but this dataset was created only for unit tests, but users see them and replicate the unit tests.

justinormont: Are you asking about how AutoML should handle the files? Or just for this example?

For AutoML in general, the files can be large and we wouldn't want to make another copy. 

For just this example, I'd recommend either a relative path to the dataset (to not duplicate the dataset in the repo), or have the example download a larger version from the CDN.

luisquintanilla: Yep. I was asking about AutoML but avoiding duplicates makes sense. Thanks.

luisquintanilla: Added data to console app and copied it to build directory. Usually, this would not be the suggested approach because both the console app and model class library are auto-generated but this was done for the purposes of this sample to be able to run after being cloned.

luisquintanilla: Tried working with larger 160k dataset. However, the data seems to have a larger number non-toxic than toxic samples. When used in the app, it doesn't return the expected result. Using the yelp dataset yielded better results. The data is perfectly balanced, although at the expense of a smaller dataset (1k observations).

PR #614, opened 2019-08-20
Adds Model Explainability sample

CESARDELATORRE:
Check readme.md when testing in VS (Troubleshooting section)


changhuixu:
@CESARDELATORRE 
I am not sure how to interpret the features' list in the PFI result. The screenshot below shows the console output from the `TaxiFarePrediction` project.

a

The feature column names have duplicates. Then which one is which? And the "`Label`" columns are in the list too, which doesn't make sense, at least to me. Is there a way to extract the real feature columns in the training pipeline?

I also have question about the parameter of `numberOfExamplesToUser=100`, which might be too small to allow features go over all possible values. What is the guideline to set this value?

Could you please answer my questions?
Thank you very much.

CESARDELATORRE:
Adding @codemzs and @gvashishtha since I work for a different team now.

PR #642, opened 2019-09-07
Fix broken link

ryandle:
The backslash was being Url encoded which broke the link. Changing it to a forward slash.


bamurtaugh:
LGTM, thank you!

PR #646, opened 2019-09-10
Update Model Builder Razor Sample to use Wikidetox Dataset

luisquintanilla:
Adapted sample to use Wikidetox dataset. No change in functionality, just adapted Web App & Console App for the dataset.

To Do:

- Update README
  - [x] Update screenshot
  - [x] Add more information on Model Builder
  - [x] Add instructions on how to adapt application to use Yelp dataset


luisquintanilla:
@CESARDELATORRE @bamurtaugh @rustd @briacht 

Please review when you get a chance. Thanks.

bamurtaugh:
Builds and runs correctly. 

Sorry if I had missed an explanation earlier, but why are we switching to the Wikidetox dataset (i.e. how does it make it easier for users to use the data)? I noticed we switched from positive/negative to toxic/not toxic, but I would've thought positive/negative sentiment would be a clearer point to start for new users?

luisquintanilla:
Thanks for taking a look @bamurtaugh. Originally the sample used the Yelp dataset (no header) which produces better results. At the moment, Model Builder does not handle data sources without a header which means the user has to manually set the column names. As a tab-delimited file, depending on which text editor you use, the tab character may not be represented the same so the accompanying tutorial used PowerShell to add the column names. While the step was relatively trivial, it was a break from the experience Model Builder intends for users.

bamurtaugh:
@luisquintanilla Thanks for the explanation!

I agree that it makes sense to minimize the number of steps for when the user is just starting. In this case, we can stick with the Wikidetox data since it's easier to start off with, and maybe you could add a statement somewhere in the readme linking to the Yelp dataset explaining, `"You can also try sentiment analysis with Model Builder using a data set of Yelp reviews (link here). The Yelp data set produces more accurate results, but it doesn't have a header to label each column. In order to use it, <explain adding header>..."`

If you agree this would be worthwhile, you could add that explanation in the readme and then we can merge in these changes.

samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazor/Pages/Index.cshtml.cs
```diff
@@ -25,12 +25,11 @@ public void OnGet()
 
         public IActionResult OnGetAnalyzeSentiment([FromQuery] string text)
         {
-            var input = new ModelInput { Comment = text };
+            if (String.IsNullOrEmpty(text)) return Content("Neutral");
+            var input = new ModelInput { SentimentText = text };
             var prediction = _predictionEnginePool.Predict(input);
-            var sentiment = Convert.ToBoolean(prediction.Prediction) ? "Positive" : "Negative";
+            var sentiment = Convert.ToBoolean(prediction.Prediction) ? "Toxic" : "Not Toxic";
```
justinormont: More specifically, the label is "Personal attack" vs. "Not personal attack"

samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazorML.ConsoleApp/ModelBuilder.cs
```diff
@@ -56,25 +56,15 @@ private static async Task DownloadData()
         {
             using (var client = new HttpClient())
             {
-                var response = await client.GetStreamAsync("https://archive.ics.uci.edu/ml/machine-learning-databases/00331/sentiment%20labelled%20sentences.zip");
+                var response = await client.GetStreamAsync("https://raw.githubusercontent.com/dotnet/machinelearning/master/test/data/wikipedia-detox-250-line-data.tsv");
```
justinormont: I would recommend a larger sample of the dataset. It's rather unlikely the model will be very good when trained on 250 rows. 

Background:
The 250 row sample was created for a unit test to check for equality without respect to how good a model it produces. Just as long as it produced the same model as last time, the unit test shows nothing has changed.

samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazorML.ConsoleApp/ModelBuilder.cs
```diff
@@ -83,13 +73,13 @@ private static async Task DownloadData()
...
-            var dataProcessPipeline = mlContext.Transforms.Text.FeaturizeText("Comment_tf", "Comment")
-                                      .Append(mlContext.Transforms.CopyColumns("Features", "Comment_tf"))
+            var dataProcessPipeline = mlContext.Transforms.Text.FeaturizeText("SentimentText_tf", "SentimentText")
+                                      .Append(mlContext.Transforms.CopyColumns("Features", "SentimentText_tf"))
                                       .Append(mlContext.Transforms.NormalizeMinMax("Features", "Features"))
                                       .AppendCacheCheckpoint(mlContext);
 
             // Set the training algorithm 
-            var trainer = mlContext.BinaryClassification.Trainers.SdcaLogisticRegression(labelColumnName: "Sentiment", featureColumnName: "Features");
+            var trainer = mlContext.BinaryClassification.Trainers.AveragedPerceptron(labelColumnName: "Sentiment", numberOfIterations: 10, featureColumnName: "Features");
```
justinormont: Good change. AP w/ 10 iterations is the winning trainer we have for general text classification. We've benchmarked this combination rather throughly.

samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazorML.ConsoleApp/Program.cs
```diff
@@ -45,12 +45,10 @@ static async Task Main(string[] args)
...
         private static ModelInput CreateSingleDataSample(MLContext mlContext)
         {
-
-            // Here (ModelInput object) you could provide new test data, hardcoded or from the end-user application.
+            // Here (ModelInput object) you could provide new test data, hardcoded or from the end-user application, instead of the row from the file.
             ModelInput sampleForPrediction = new ModelInput
             {
-                Comment= "Wow... Loved this place.",
-                Sentiment=true
+                SentimentText = "Model Builder is cool!"
```
justinormont: Fully agree!

samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazorML.ConsoleApp/ModelBuilder.cs
luisquintanilla: Agreed this dataset is limiting. Unfortunately given the constraints around building a plug-and-play experience, re-hosting datasets, no feature engineering in Model Builder at the moment and the 160k rows dataset being imbalanced (90% non-toxic, 10% toxic), this is the "best" dataset for us to work with at the moment considering it's already part of the [getting started experience](https://dotnet.microsoft.com/learn/ml-dotnet/get-started-tutorial/intro) on the .NET site . However, if there is another version of the version of the dataset that can be consumed as-is with no manipulation from the user, I can switch the dataset then.

samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazorML.ConsoleApp/ModelBuilder.cs
luisquintanilla: That's a kudos to the AutoML / Model Builder teams since this is part of the code-gen. I take no credit for that 🙂

luisquintanilla:
@bamurtaugh I'll add that section to additional changes I'll be making to the README.

luisquintanilla:
Thanks for the review @bamurtaugh

 :shipit: ?

bamurtaugh:
LGTM, going to merge now :shipit:

samples/modelbuilder/BinaryClassification_Sentiment_Razor/SentimentRazor/Startup.cs
```diff
@@ -39,7 +39,8 @@ public void ConfigureServices(IServiceCollection services)
             
             services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_2);
 
-            services.AddPredictionEnginePool<ModelInput, ModelOutput>().FromFile(_modelPath);
+            services.AddPredictionEnginePool<ModelInput, ModelOutput>()
+                    .FromFile(_modelPath);
```
Youssef1313: This change has broken the docs, the line was referenced by its number.
I've opened https://github.com/dotnet/docs/pull/14821 to fix that in the docs repo.

PR #660, opened 2019-09-19
Update Nuget packages and fix reference paths

drlawler:
I opened the solution in VS2019, and quite a few references didn't load correctly.  It appears the relative paths were not quite right.  I got all projects to build, as well as updated Nuget packages to the most recent available.

### DIFF
--- a/samples/csharp/end-to-end-apps/AnomalyDetection-Sales/SpikeDetectionE2EApp/SpikeDetection.WinForms/SpikeDetection.WinForms.csproj
+++ b/samples/csharp/end-to-end-apps/AnomalyDetection-Sales/SpikeDetectionE2EApp/SpikeDetection.WinForms/SpikeDetection.WinForms.csproj
@@ -118,31 +118,31 @@
       <Version>$(MicrosoftMLVersion)</Version>
     </PackageReference>
     <PackageReference Include="Newtonsoft.Json">
-      <Version>10.0.3</Version>
+      <Version>12.0.2</Version>
     </PackageReference>
     <PackageReference Include="System.Buffers">
-      <Version>4.4.0</Version>
+      <Version>4.5.0</Version>
     </PackageReference>
     <PackageReference Include="System.CodeDom">
-      <Version>4.4.0</Version>
+      <Version>4.5.0</Version>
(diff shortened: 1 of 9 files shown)


drlawler:
@bamurtaugh that wasn't a file that I changed.  But looking at it, I believe it is correct to have the TaxiFarePredictionConsoleApp included in the path.  The next line adds the "inputs" folder, and without the "TaxiFarePredictionConsoleApp" being included in the path, it would be looking for "inputs" under the main TaxiFarePrediction folder.  

I'm not sure why this doesn't work on yours.  

Also, curious how you found the issue so quick.  is there a "Run all apps" or something like that?  I have just started digging into the solution, and just wanted to share back the changes I needed to make to get it running.

bamurtaugh:
@drlawler Yes I saw you hadn't changed that line, but I had to make that change in order for the app to work when I was testing it just now. But since yours runs correctly as-is, I'll have to investigate this further. I found that issue rather quickly since I just went through to test the projects you had made changes to in your PR. 

In order to run the apps more quickly (as opposed to opening each individual solution, running it, and then going back to the list of solutions to select the next one), you can select: samples -> csharp -> v1.3.1-All-Samples.sln (I'm not sure if this was the way you were already running them though?).

As you had mentioned you've just started digging into the solution, is this PR still a work in progress? Also, for the updates you made to the various nuget packages, did you need to update those for the projects to run on your machine? Or did you just want to update them as a good practice? I ask because I am able to run the projects as-is without updating those packages, but you've definitely done some great updating work here. 😄 

If you have any other questions, please let us know!

drlawler:
@bamurtaugh This PR was as far as I intended to take it.  It really was just about updating packages and fixing package references.  As far as the relative paths in the code, I didn't intend to change any of the code, or the locations of the data files.  I only changed the project and package files.

If there is something else I can do to help you validate this, please let me know.

CESARDELATORRE:
Closing this PR. We prefer to take PRs targeting single projects, so each PR targeting a single project so it's easier to track issues or regression bugs.
This PR targets too many different projects. Please, if you want to provide PR, create each PR targeting a single sample app. thanks! 👍

Issue #684, opened 2019-10-09
Save the model to disk earlier in DeepLearning_ImageClassification_Training project


jeremylindsayni:
Hi - I've been working with one of the sample projects (DeepLearning_ImageClassification_Training) - these projects are great and super helpful for getting started. But I've found something that caused me an issue. 

At a high level, the current flow of the example is:
- Generate and evaluate the model (running this takes a long time),
- Try the model with a single prediction,
- Save the model to disk.

_Steps to recreate:_
I hit an exception when I ran the "TrySinglePrediction" method - root cause of this was my system and my fault, not the code's (I was trying to use my own images). But this exception meant I had to stop the process, and at the time of stopping the process, the model hadn't been saved to disk yet. This also meant I had to re-run the entire time-consuming process to re-generate and evaluate the model - which was exactly the same model as the first time I ran the code.

_Proposed behaviour:_
I think it would be better to save the model to disk as soon as possible after generating and evaluating it, and then testing the model with a single prediction. This test doesn't change the model at all, and if it fails for whatever reason, at least the model zip file is already saved to disk - so I can skip straight to the "ImageClassification.Predict" project, and don't have to wait for the model zip-file to be generated again.

So the change to the flow of the example would turn it into:
- Generate and evaluate the model,
- Save the model to disk,
- Try the model with a single prediction.

I will submit a pull request for this change - I hope it is helpful.

CESARDELATORRE:
Thanks for this suggestion. I think it's a good idea to change the flow as you mentioned.
However, I'm already working in parallel on a new version with the evolved API in 1.4-Preview-2 so I'll probably simply make that suggestion in the sample update instead of taking this PR since the code is partially different due to the support for in-memory images. So it'll be easier to simply change my newer code with this suggestion, ok? :)

I'll update this new sample code (using 1.4-Preview-2 and in-memory images for scoring) in the repo this week. today/tomorrow, hopefully. 👍

jeremylindsayni:
Hey Cesar - thank you for looking at this so soon. What you suggest sounds perfect - I'll close my pull request. I can leave this issue open if it's helpful to refer to it when you make your changes to 1.4-Preview-2, or I'm happy to close it if you are. Thanks again 👍

CESARDELATORRE:
Cool! Let's keep the issue open until I publish the sample so I don't forget about this minor change.
You can close the PR, though.
Thanks! 👍

luisquintanilla:
[...]

PR #685, opened 2019-10-09
Save the model to disk earlier - address issue #684

jeremylindsayni:
This pull request addresses issue #684 - the model is now saved to disk before it's tested with images that aren't in the training set. This means that if there's any exception during that test, we won't all the work done in model generation/evaluation because it has been saved to disk already, and can be re-used.

Hope this helps,
Jeremy


jeremylindsayni:
Closing as this change is expected in 1.4-Preview-2, as per discussion with @CESARDELATORRE in original issue.

PR #725, opened 2019-10-29
Update Azure Functions sample to run on Azure

luisquintanilla:
Updated sample to run on Azure

- Depending on the environment (Dev=Local/Prod=Azure), set the path in the `Startup` class where the model is stored.


luisquintanilla:
@CESARDELATORRE / @bamurtaugh ready for review when you get a chance. Thanks.

PR #726, opened 2019-10-30
Uploading to Master New Text Sample

MaxAkbar:
Added a new sample on how to cluster text.

Sorry forgot to mention that the sample data came from https://www.kaggle.com/rmisra/news-category-dataset but it's a subset of it and wanted to keep it smaller.

This addresses: dotnet/machinelearning#4106

PR #730, opened 2019-10-30
Revert "Uploading to Master New Text Sample"

CESARDELATORRE:
Reverts dotnet/machinelearning-samples#726

Reverting this merge for this new sample. I found a few issues:
- This file shouldn't be part of this sample PR:
https://github.com/dotnet/machinelearning-samples/blob/master/samples/csharp/examples/DeepLearning_TensorFlowMLNETInceptionv3ModelScoring/TensorFlowMLNETInceptionv3ModelScoring.csproj
- ClusteringNewsArticles.Perdict (Rename to Predict)
- Why a so high number of clusters? (41 clusters), is that intended to be that high?

Please, update the PR after we upgrade to v1.4 GA which is almost happening now..
Thanks! :)


MaxAkbar:
Sorry didn't see that file, I will take care of it and resubmit. 
As for the number of clusters, I did a count of different categories and it came to be 41. I will do some more work to see if I can lower it.
I will wait for v1.4 thank you for the feedback :).

CESARDELATORRE:
Thanks to you for contributing with such an interesting sample! :)

PR #737, opened 2019-11-05
Web Ranking sample

nicolehaugen:
Updated web ranking sample to 1.4 - this fixes the null reference crash that was happening now that we no longer need to use reflection for getting evaluation metrics for different NDCG values.

### DIFF
--- a/samples/csharp/getting-started/Ranking_Web/WebRanking/WebRanking.csproj
+++ b/samples/csharp/getting-started/Ranking_Web/WebRanking/WebRanking.csproj
@@ -6,9 +6,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.ML" Version="1.4.0-preview-28021-0" />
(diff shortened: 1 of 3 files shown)


luisquintanilla:
Hi @nicolehaugen,

Thanks for the update. Looks good to me. I see there are some merge conflicts. Once those are resolved, I'll merge.

samples/csharp/getting-started/Ranking_Web/WebRanking/WebRanking.csproj
```diff
@@ -6,8 +6,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.ML" Version="$(MicrosoftMLVersion)" />
-    <PackageReference Include="Microsoft.ML.LightGbm" Version="$(MicrosoftMLVersion)" />
+    <PackageReference Include="Microsoft.ML" Version="1.4.0" />
+    <PackageReference Include="Microsoft.ML.LightGbm" Version="1.4.0" />
```
justinormont: Why move from `$(MicrosoftMLVersion)`?

PR #761, opened 2019-12-27
Cleanup ModelScoringTester.fs

forki:


samples/fsharp/getting-started/Regression_BikeSharingDemand/BikeSharingDemand/BikeSharingDemandConsoleApp/ModelScoringTester.fs
```diff
@@ -6,37 +6,38 @@ open DataStructures
...
-        Temperature = Single.Parse(x.[10], CultureInfo.InvariantCulture)
-        NormalizedTemperature = Single.Parse(x.[11], CultureInfo.InvariantCulture)
-        Humidity = Single.Parse(x.[12], CultureInfo.InvariantCulture)
-        Windspeed = Single.Parse(x.[13], CultureInfo.InvariantCulture)
-        Count = Single.Parse(x.[16], CultureInfo.InvariantCulture)
-    })
+    |> Seq.filter (fun x -> not (String.IsNullOrEmpty x))
+    |> Seq.map (fun x ->
+        let fields = x.Split(',')
+        { Season = parseSingle fields.[2]
```
kevmal: maybe just drop the `parseSingle` def and use `float32`, its basically equivalent.

forki: I don't think that's using InvariantCulture.

kevmal: It's using InvariantCulture, the difference is `NumberStyles.AllowThousands` is not set in favor of striping all underscores prior to parsing https://github.com/dotnet/fsharp/blob/f9ea5824a321a0df73675ad392458af79297f092/src/fsharp/FSharp.Core/prim-types.fs#L3578

PR #852, opened 2020-09-24
Add Ranking AutoML Sample

jwood803:
Add Ranking sample for AutoML.

Update for [this issue](https://github.com/dotnet/machinelearning/issues/5408)

### DIFF
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -100,7 +100,7 @@ namespace Common
 
         internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
         {
-            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.DiscountedCumulativeGains.Average() ?? double.NaN,14:F4} {metrics?.NormalizedDiscountedCumulativeGains.Average() ?? double.NaN,14:F2} {runtimeInSeconds.Value,9:F1}", Width);
+            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.DiscountedCumulativeGains.Average() ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains.Average() ?? double.NaN,9:F2} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
         internal static void PrintIterationException(Exception ex)
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -125,7 +125,7 @@ namespace Common
 
         internal static void PrintRankingMetricsHeader()
         {
-            CreateRow($"{"",-4} {"Trainer",-14}, {"nDCG@1",9}, {"nDCG@2",9}, {"nDCG@10",9}, {"DCG@1",9}, {"Duration",9}", Width);
+            CreateRow($"{"",-4} {"Trainer",-14}, {"nDCG@1",9}, {"nDCG@3",9}, {"nDCG@10",9}, {"DCG@10",9}, {"Duration",9}", Width);
         }
 
         private static void CreateRow(string message, int width)
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -100,7 +100,7 @@ namespace Common
 
         internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
         {
-            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[1] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[0] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
+            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
         internal static void PrintIterationException(Exception ex)
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -60,8 +60,8 @@ namespace Common
             Console.WriteLine($"************************************************************");
             Console.WriteLine($"*    Metrics for {name} ranking model   ");
             Console.WriteLine($"*-----------------------------------------------------------");
-            Console.WriteLine($"    Discounted Cumulative Gains = {metrics.DiscountedCumulativeGains.Average():0.####}");
-            Console.WriteLine($"    Normalized Discounted Cumulative Gains = {metrics.NormalizedDiscountedCumulativeGains.Average():0.####}, a value between 0 and 1, the closer to 1, the better");
+            Console.WriteLine($"    Discounted Cumulative Gains = {metrics.DiscountedCumulativeGains[0]:0.####}");
+            Console.WriteLine($"    Normalized Discounted Cumulative Gains = {metrics.NormalizedDiscountedCumulativeGains[0]:0.####}, a value between 0 and 1, the closer to 1, the better");
         }
 
         public static void ShowDataViewInConsole(MLContext mlContext, IDataView dataView, int numberOfRows = 4)
@@ -100,7 +100,7 @@ namespace Common
 
         internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
         {
-            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4}  {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
+            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[0] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
         internal static void PrintIterationException(Exception ex)
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -44,10 +44,10 @@ namespace Common
         public static void PrintMulticlassClassificationMetrics(string name, MulticlassClassificationMetrics metrics)
         {
             Console.WriteLine($"************************************************************");
-            Console.WriteLine($"*    Metrics for {name} multi-class classification model   ");
+            Console.WriteLine($"*    Metrics for {name} multi-class classification model    ");
             Console.WriteLine($"*-----------------------------------------------------------");
-            Console.WriteLine($"    MacroAccuracy = {metrics.MacroAccuracy:0.####}, a value between 0 and 1, the closer to 1, the better");
-            Console.WriteLine($"    MicroAccuracy = {metrics.MicroAccuracy:0.####}, a value between 0 and 1, the closer to 1, the better");
+            Console.WriteLine($"    MacroAccuracy = {metrics.MacroAccuracy:0.####}, a value from 0 and 1, where closer to 1.0 is better");
+            Console.WriteLine($"    MicroAccuracy = {metrics.MicroAccuracy:0.####}, a value from 0 and 1, where closer to 1.0 is better");
             Console.WriteLine($"    LogLoss = {metrics.LogLoss:0.####}, the closer to 0, the better");
             Console.WriteLine($"    LogLoss for class 1 = {metrics.PerClassLogLoss[0]:0.####}, the closer to 0, the better");
             Console.WriteLine($"    LogLoss for class 2 = {metrics.PerClassLogLoss[1]:0.####}, the closer to 0, the better");
@@ -55,13 +55,13 @@ namespace Common
             Console.WriteLine($"************************************************************");
         }
 
-        public static void PrintRankingMetrics(string name, RankingMetrics metrics)
+        public static void PrintRankingMetrics(string name, RankingMetrics metrics, uint optimizationMetricTruncationLevel)
         {
             Console.WriteLine($"************************************************************");
-            Console.WriteLine($"*    Metrics for {name} ranking model   ");
+            Console.WriteLine($"*             Metrics for {name} ranking model              ");
             Console.WriteLine($"*-----------------------------------------------------------");
-            Console.WriteLine($"    Discounted Cumulative Gain (DCG@10) = {metrics?.DiscountedCumulativeGains?[9] ?? double.NaN:0.####}");
-            Console.WriteLine($"    Normalized Discounted Cumulative Gain (NDCG@10) = {metrics?.NormalizedDiscountedCumulativeGains?[9] ?? double.NaN:0.####}, a value between 0 and 1, the closer to 1, the better");
+            Console.WriteLine($"    Normalized Discounted Cumulative Gain (NDCG@{optimizationMetricTruncationLevel}) = {metrics?.NormalizedDiscountedCumulativeGains?[(int)optimizationMetricTruncationLevel - 1] ?? double.NaN:0.####}, a value from 0 and 1, where closer to 1.0 is better");
+            Console.WriteLine($"    Discounted Cumulative Gain (DCG@{optimizationMetricTruncationLevel}) = {metrics?.DiscountedCumulativeGains?[(int)optimizationMetricTruncationLevel - 1] ?? double.NaN:0.####}");
         }
 
         public static void ShowDataViewInConsole(MLContext mlContext, IDataView dataView, int numberOfRows = 4)
@@ -100,7 +100,7 @@ namespace Common
 
         internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
         {
-            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
+            CreateRow($"{iteration,-4} {trainerName,-15} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
         internal static void PrintIterationException(Exception ex)
@@ -125,7 +125,7 @@ namespace Common
 
         internal static void PrintRankingMetricsHeader()
         {
-            CreateRow($"{"",-4} {"Trainer",-14}, {"nDCG@1",9}, {"nDCG@3",9}, {"nDCG@10",9}, {"DCG@10",9}, {"Duration",9}", Width);
+            CreateRow($"{"",-4} {"Trainer",-15} {"NDCG@1",9} {"NDCG@3",9} {"NDCG@10",9} {"DCG@10",9} {"Duration",9}", Width);
         }
 
         private static void CreateRow(string message, int width)
@@ -258,10 +258,10 @@ namespace Common
 
                 tableRows.Add(new[]
                 {
-                columnName,
-                GetColumnDataType(columnName),
-                columnPurpose
-            });
+                    columnName,
+                    GetColumnDataType(columnName),
+                    columnPurpose
+                });
             }
 
             private void AppendTableRows(ICollection<string[]> tableRows,
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -55,6 +55,15 @@ namespace Common
             Console.WriteLine($"************************************************************");
         }
 
+        public static void PrintRankingMetrics(string name, RankingMetrics metrics)
+        {
+            Console.WriteLine($"************************************************************");
+            Console.WriteLine($"*    Metrics for {name} ranking model   ");
+            Console.WriteLine($"*-----------------------------------------------------------");
+            Console.WriteLine($"    Discounted Cumulative Gains = {metrics.DiscountedCumulativeGains.Average():0.####}");
+            Console.WriteLine($"    Normalized Discounted Cumulative Gains = {metrics.NormalizedDiscountedCumulativeGains.Average():0.####}, a value between 0 and 1, the closer to 1, the better");
+        }
+
         public static void ShowDataViewInConsole(MLContext mlContext, IDataView dataView, int numberOfRows = 4)
         {
             string msg = string.Format("Show data in DataView: Showing {0} rows with the columns", numberOfRows.ToString());
@@ -89,6 +98,11 @@ namespace Common
             CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.RSquared ?? double.NaN,8:F4} {metrics?.MeanAbsoluteError ?? double.NaN,13:F2} {metrics?.MeanSquaredError ?? double.NaN,12:F2} {metrics?.RootMeanSquaredError ?? double.NaN,8:F2} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
+        internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
+        {
+            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.DiscountedCumulativeGains.Average() ?? double.NaN,14:F4} {metrics?.NormalizedDiscountedCumulativeGains.Average() ?? double.NaN,14:F2} {runtimeInSeconds.Value,9:F1}", Width);
+        }
+
         internal static void PrintIterationException(Exception ex)
         {
             Console.WriteLine($"Exception during AutoML iteration: {ex}");
@@ -107,6 +121,11 @@ namespace Common
         internal static void PrintRegressionMetricsHeader()
         {
             CreateRow($"{"",-4} {"Trainer",-35} {"RSquared",8} {"Absolute-loss",13} {"Squared-loss",12} {"RMS-loss",8} {"Duration",9}", Width);
+        }
+
+        internal static void PrintRankingMetricsHeader()
+        {
+            CreateRow($"{"",-4} {"Trainer",-35}, {"DiscountedCumulativeGains",14}, {"NoramlizedDiscountedCumuliativeGains",14}", Width);
         }
 
         private static void CreateRow(string message, int width)
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -100,7 +100,7 @@ namespace Common
 
         internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
         {
-            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[0] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
+            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[1] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[0] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
         internal static void PrintIterationException(Exception ex)
@@ -125,7 +125,7 @@ namespace Common
 
         internal static void PrintRankingMetricsHeader()
         {
-            CreateRow($"{"",-4} {"Trainer",-35}, {"NoramlizedDiscountedCumuliativeGains@1",9}, {"NoramlizedDiscountedCumuliativeGains@3",9} {"DiscountedCumuliativeGains@1",9}", Width);
+            CreateRow($"{"",-4} {"Trainer",-14}, {"nDCG@1",9}, {"nDCG@2",9}, {"nDCG@10",9}, {"DCG@1",9}, {"Duration",9}", Width);
         }
 
         private static void CreateRow(string message, int width)
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -60,8 +60,8 @@ namespace Common
             Console.WriteLine($"************************************************************");
             Console.WriteLine($"*    Metrics for {name} ranking model   ");
             Console.WriteLine($"*-----------------------------------------------------------");
-            Console.WriteLine($"    Discounted Cumulative Gains = {metrics.DiscountedCumulativeGains[0]:0.####}");
-            Console.WriteLine($"    Normalized Discounted Cumulative Gains = {metrics.NormalizedDiscountedCumulativeGains[0]:0.####}, a value between 0 and 1, the closer to 1, the better");
+            Console.WriteLine($"    Discounted Cumulative Gain (DCG@10) = {metrics?.DiscountedCumulativeGains?[9] ?? double.NaN:0.####}");
+            Console.WriteLine($"    Normalized Discounted Cumulative Gain (NDCG@10) = {metrics?.NormalizedDiscountedCumulativeGains?[9] ?? double.NaN:0.####}, a value between 0 and 1, the closer to 1, the better");
         }
 
         public static void ShowDataViewInConsole(MLContext mlContext, IDataView dataView, int numberOfRows = 4)
@@ -100,7 +100,7 @@ namespace Common
 
         internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
         {
-            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[0] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
+            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[0] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
         internal static void PrintIterationException(Exception ex)
@@ -125,7 +125,7 @@ namespace Common
 
         internal static void PrintRankingMetricsHeader()
         {
-            CreateRow($"{"",-4} {"Trainer",-35}, {"DiscountedCumulativeGains",14}, {"NoramlizedDiscountedCumuliativeGains",14}", Width);
+            CreateRow($"{"",-4} {"Trainer",-35}, {"NoramlizedDiscountedCumuliativeGains@1",9}, {"NoramlizedDiscountedCumuliativeGains@3",9} {"DiscountedCumuliativeGains@1",9}", Width);
         }
 
         private static void CreateRow(string message, int width)
--- a/samples/csharp/common/AutoML/ConsoleHelper.cs
+++ b/samples/csharp/common/AutoML/ConsoleHelper.cs
@@ -100,7 +100,7 @@ namespace Common
 
         internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
         {
-            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.DiscountedCumulativeGains.Average() ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains.Average() ?? double.NaN,9:F2} {runtimeInSeconds.Value,9:F1}", Width);
+            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4}  {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
         internal static void PrintIterationException(Exception ex)
--- a/samples/csharp/getting-started/Ranking_AutoML/Ranking/README.md
+++ b/samples/csharp/getting-started/Ranking_AutoML/Ranking/README.md
@@ -10,7 +10,7 @@ The ability to perform ranking is a common problem faced by search engines since
 
 Ranking is useful to any scenario where it is important to list items in an order that increases the likelihood of a click, purchase, reservation, etc.
  
-In this sample, we show how to apply ranking to search engine results. To perform ranking, there are two algorithms currently available - FastTree Boosting (FastRank) and Light Gradient Boosting Machine (LightGBM). We use the LightGBM's LambdaRank implementation in this sample to automatically build an ML model to predict ranking. 
+In this sample, we show how to apply ranking to search engine results. AutoML sweeps over the models and associated hyperparameters to optimize [NDCG@N](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.rankingmetrics.normalizeddiscountedcumulativegains?view=ml-dotnet) (normalized discounted cumulative gains) or [DCG@N](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.rankingmetrics.discountedcumulativegains?view=ml-dotnet) (discounted cumulative gains).
 
 ## Dataset
 The data used by this sample is based on a public [dataset provided by Microsoft](https://www.microsoft.com/en-us/research/project/mslr/) originally provided Microsoft Bing. The dataset is released under a [CC-by 4.0](https://creativecommons.org/licenses/by/4.0/) license and includes training, validation, and testing data.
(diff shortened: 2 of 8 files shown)


jwood803:
@justinormont Feel free to let me know if anything is missing or should be changed with this sample.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,175 @@
...
+            // invoke after each model it produces and evaluates.
+            var progressHandler = new RankingExperimentProgressHandler();
+
+            // STEP 4: Run AutoML ranking experiment
+            ConsoleHelper.ConsoleWriteHeader("=============== Running AutoML experiment ===============");
+            Console.WriteLine($"Running AutoML ranking experiment for {ExperimentTime} seconds...");
+
+            ExperimentResult<RankingMetrics> experimentResult = mlContext.Auto()
+                .CreateRankingExperiment(ExperimentTime)
+                .Execute(trainDataView, progressHandler: progressHandler);
```
justinormont, line 67: Since it's available, I'd recommend handing it both the train and validation sets. 

When given a singular training set with out a validation set, it will split the training datasets (or use CV) to auto-create a validation set. But since we have one we should use it.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            // Get a row for single prediction
+            sampleRankingData = mlContext.Data.CreateEnumerable<RankingData>(mlContext.Data.TakeRows(testDataView, 1), reuseRowObject: false)
+                .FirstOrDefault();
+
+            var metrics = mlContext.Ranking.Evaluate(predictions);
+
+            ConsoleHelper.PrintRankingMetrics(bestRun.TrainerName, metrics);
+
+            // STEP 6: Save/persist the trained model to a .ZIP file
```
justinormont, line 88: If you're up for it, you could demonstrate thrice training. 

The general idea is to progressively refit the best found pipeline with merged data from all of the datasets in a three step process.

step | fitting data | scoring data | notes
-----|------|---|----
1 | train | valid | Done within AutoML to find the best pipeline
2 | train+valid | test | Gives final metrics which estimate the model's perf in production
3 | train+valid+test | N/A | Gives final model to launch to production

I have an example created for another [bug](https://github.com/dotnet/machinelearning/issues/5320#issuecomment-662940586): 
full example -- https://dotnetfiddle.net/nWpCkP
```c#
// Re-fit best pipeline on train and validation data, to produce 
// a model that is trained on as much data as is available while
// still having test data for the final estimate of how well the
// model will do in production.
Console.WriteLine("\n===== Refitting on train+valid and evaluating model's rsquared with test data =====");
var TrainPlusValidationDataView = textLoader.Load(new MultiFileSource(TrainDataPath, ValidationDataPath));
var refitModel1 = experimentResult.BestRun.Estimator.Fit(TrainPlusValidationDataView);
IDataView predictionsRefitOnTrainPlusValidation = refitModel1.Transform(TestDataView);
var metricsRefitOnTrainPlusValidation = mlContext.Regression.Evaluate(predictionsRefitOnTrainPlusValidation, labelColumnName: "Label", scoreColumnName: "Score");
Console.WriteLine("|" + $"{"-",-4} {experimentResult.BestRun.TrainerName,-35} {metricsRefitOnTrainPlusValidation?.RSquared ?? double.NaN,8:F4} {metricsRefitOnTrainPlusValidation?.MeanAbsoluteError ?? double.NaN,13:F2} {metricsRefitOnTrainPlusValidation?.MeanSquaredError ?? double.NaN,12:F2} {metricsRefitOnTrainPlusValidation?.RootMeanSquaredError ?? double.NaN,8:F2} {"-",9}".PadRight(112) + "|");

// Re-fit best pipeline on train, validation, and test data, to 
// produce a model that is trained on as much data as is available.
// This is the final model that can be deployed to production.
Console.WriteLine("\n===== Refitting on train+valid+test to get the final model to launch to production =====");
var TrainPlusValidationPlusTestDataView = textLoader.Load(new MultiFileSource(TrainDataPath, ValidationDataPath, TestDataPath));
var refitModel2 = experimentResult.BestRun.Estimator.Fit(TrainPlusValidationPlusTestDataView);
```

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,175 @@
...
+                    client.DownloadFile(testDatasetUrl, testDatasetPath);
+                }
+            }
+
+            Console.WriteLine("===== Download is finished =====\n");
+        }
+
+        private static void PrintTopModels(ExperimentResult<RankingMetrics> experimentResult)
+        {
+            // Get top few runs ranked by accuracy
```
justinormont, line 161:
```suggestion
            // Get top few runs ranked by NDCG
```

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,175 @@
...
+            var topRuns = experimentResult.RunDetails
+                .Where(r => r.ValidationMetrics != null && !double.IsNaN(r.ValidationMetrics.NormalizedDiscountedCumulativeGains.Average()))
+                .OrderByDescending(r => r.ValidationMetrics.NormalizedDiscountedCumulativeGains.Average()).Take(3);
+
+            Console.WriteLine("Top models ranked by accuracy --");
```
justinormont, line 166:
```suggestion
            Console.WriteLine("Top models ranked by NDCG --");
```

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,175 @@
...
+            ConsoleHelper.ConsoleWriteHeader("=============== Testing prediction engine ===============");
+
+
+            ITransformer trainedModel = mlContext.Model.Load(ModelPath, out var modelInputSchema);
+            Console.WriteLine($"=============== Loaded Model OK  ===============");
+
+            // Create prediction engine related to the loaded trained model
+            var predictionEngine = mlContext.Model.CreatePredictionEngine<RankingData, RankingPrediction>(trainedModel, inputSchema: modelInputSchema);
+
+            var predictionResult = predictionEngine.Predict(sampleRankingData);
```
justinormont, line 107: Showing off ranked results of a single query would be more beneficial. 

I think this is done as:
* Get the testset's first row's `GroupID`
* Filter the IDataView to rows with that `GroupID` <br/>(or just know the number of rows in the first group)
* Score those rows
* Sort the rows by `score` (descending)
* Display rows, now in ranked order

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -89,6 +98,11 @@ internal static void PrintIterationMetrics(int iteration, string trainerName, Re
             CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.RSquared ?? double.NaN,8:F4} {metrics?.MeanAbsoluteError ?? double.NaN,13:F2} {metrics?.MeanSquaredError ?? double.NaN,12:F2} {metrics?.RootMeanSquaredError ?? double.NaN,8:F2} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
+        internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
+        {
+            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.DiscountedCumulativeGains.Average() ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains.Average() ?? double.NaN,9:F2} {runtimeInSeconds.Value,9:F1}", Width);
```
justinormont, line 103: I would print the NDCG@k for k=1,3,10 and DCG@k for k=10
```suggestion
            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
```

These represent taking into account only the top K results, and ignoring the quality of the results beyond there. https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
jwood803: This is an interesting concept! I'll definitely give this a go.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
justinormont: Thanks.

samples/csharp/common/AutoML/ConsoleHelper.cs
justinormont: Thanks for updating. 

I think you'll have to update the printed header row to match the newly added metrics. It should be approximately, "NDCG@1  NDCG@3 NDCG@10 DCG@10 RunTime" (with the right formatting of course).

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -55,6 +55,15 @@ public static void PrintMulticlassClassificationMetrics(string name, MulticlassC
             Console.WriteLine($"************************************************************");
         }
 
+        public static void PrintRankingMetrics(string name, RankingMetrics metrics)
+        {
+            Console.WriteLine($"************************************************************");
+            Console.WriteLine($"*    Metrics for {name} ranking model   ");
+            Console.WriteLine($"*-----------------------------------------------------------");
+            Console.WriteLine($"    Discounted Cumulative Gains = {metrics.DiscountedCumulativeGains.Average():0.####}");
+            Console.WriteLine($"    Normalized Discounted Cumulative Gains = {metrics.NormalizedDiscountedCumulativeGains.Average():0.####}, a value between 0 and 1, the closer to 1, the better");
```
justinormont, line 64: There's a few places still using the `.Average()`, which should print one of the values of the array. Another choice is to print all.

```suggestion
            Console.WriteLine($"    Normalized Discounted Cumulative Gain (NDCG@10) = {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN:0.####}, a value between 0 and 1, the closer to 1, the better");
```

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -89,6 +98,11 @@ internal static void PrintIterationMetrics(int iteration, string trainerName, Re
             CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.RSquared ?? double.NaN,8:F4} {metrics?.MeanAbsoluteError ?? double.NaN,13:F2} {metrics?.MeanSquaredError ?? double.NaN,12:F2} {metrics?.RootMeanSquaredError ?? double.NaN,8:F2} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
+        internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
+        {
+            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4}  {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
```
justinormont, line 103: What is the length of the array? This will be max `k` in `NDCG@k`, and likely also the default set in the AutoML code for `k` for which NDCG value to optimize (the optimization metric).

Is there an input for `k` in the AutoML API? If so, it would be good to set to perhaps `10`, and ensure the user's specified `k` value is within the printed metrics. 

Setting a value in a sample has a good pedagogical value as it lets users know they should be setting this value. The input would control which `k` in `NDCG@k` that AutoML should optimize towards (aka, the optimization metric).

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
jwood803: Thanks for letting me know about this! I think a video on this would be good to make. 😄

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
jwood803: I think that's what they do [here](https://github.com/dotnet/machinelearning-samples/blob/master/samples/csharp/getting-started/Ranking_Web/WebRanking/Program.cs#L193) as well, so I used that as an example. Let me know if it's not correct, though.

samples/csharp/common/AutoML/ConsoleHelper.cs
jwood803: Unfortunately, for the progress handler, I'm unable to change the number of `k`. For the `Evaluate` method, however, I can specify it.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,232 @@
...
+
+            Console.WriteLine("===== Download is finished =====\n");
+        }
+
+        private static void PrintTopModels(ExperimentResult<RankingMetrics> experimentResult)
+        {
+            // Get top few runs ranked by nDCG
+            var topRuns = experimentResult.RunDetails
+                .Where(r => r.ValidationMetrics != null && !double.IsNaN(r.ValidationMetrics.NormalizedDiscountedCumulativeGains[0]))
+                .OrderByDescending(r => r.ValidationMetrics.NormalizedDiscountedCumulativeGains[0]).Take(3);
```
justinormont, line 221: Looking at the AutoML code, it defaults to NDCG@10 (or the last element of the array):
https://github.com/dotnet/machinelearning/blob/f87a3bbd8adb12934dae0a0060813ce9b7500664/src/Microsoft.ML.AutoML/Experiment/MetricsAgents/RankingMetricsAgent.cs#L31-L32

Should use the same sorting logic here. This will ensure that what the AutoML code returns as its `experimentResult.BestRun` will show sorted to the top of this list.

---

As an aside the AutoML code isn't passing through a `k` parameter for the optimization metric. This is room for improvement in our AutoML ranking API/code. Currently the AutoML code is hard-coded to optimize NDCG@10 (or DCG@10) instead of a user supplied `k` in `NDCG@k`. Having the user supply this parameter instead of hard coding lets the AutoML code optimize the model towards the user's needs; for instance NDCG@3 is useful for an auto-suggest showing three results, NDCG@10 is useful for the first page of search results, NDCG@30 optimizes the first three pages.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
justinormont: Quite so. The other ranking sample already had exactly what I was thinking of. Thanks for locating it and reusing here.

#RESOVLED

samples/csharp/common/AutoML/ConsoleHelper.cs
justinormont: Yes, you're right; it's not exposed. I left a comment on the lack of the `k` parameter for the AutoML ranking experiment settings. 

https://github.com/dotnet/machinelearning-samples/pull/852#discussion_r498425201
> As an aside the AutoML code isn't passing through a `k` parameter for the optimization metric. This is room for improvement in our AutoML ranking API/code. Currently the AutoML code is hard-coded to optimize NDCG@10 (or DCG@10) instead of a user supplied `k` in `NDCG@k`. Having the user supply this parameter instead of hard coding lets the AutoML code optimize the model towards the user's needs; for instance NDCG@3 is useful for an auto-suggest showing three results, NDCG@10 is useful for the first page of search results, NDCG@30 optimizes the first three pages.
>
> TLDR; an `@k` parameter should be exposed in the AutoML ranking experiment settings.

#RESOLVED

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
justinormont: This would make a great video. This process is most beneficial when the training set is small, or the dataset is split by time. 

Ranking datasets are often split by time with the older data in the training split, newer data in the validation, and newest in the test dataset split. There's mainly two gains in using time splits; (1) removing leakage -- see "time leakage" in https://en.wikipedia.org/wiki/Leakage_(machine_learning), and (2) the most valuable data is the most up to date data as user trends shift over time.

The newest data is the most representative of the live production traffic the model will see, hence it's a better measure of the model's performance when launched in production (aka, why we use it as the final metrics), and its use in refitting the model will create a model which performs better on the production traffic.

samples/csharp/common/AutoML/ConsoleHelper.cs
justinormont: Thanks for updating the the headers.

#RESOLVED

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -100,7 +100,7 @@ internal static void PrintIterationMetrics(int iteration, string trainerName, Re
 
         internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
         {
-            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4}  {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
+            CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[0] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
```
justinormont: I see you took out `NDCG@10`. Are the printed metrics too wide on the screen? 

If the space is important, the header could be reduced the from full `NormalizedDiscountedCumulativeGain@10` to `NDCG@10`. 

The AutoML code is optimizing `NDCG@10` / `DCG@10` ([comment](https://github.com/dotnet/machinelearning-samples/pull/852#discussion_r498425201)), hence it's beneficial to print that specific `@k` value.

jwood803: There were only three items. But since the third item is the 10th, I'll include that. Thanks for catching this!

jwood803: Did some improvements here and in the header.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
jwood803: I'll put an issue for this today to help keep track of this. I may even try to implement it 😄

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,232 @@
...
+            {
+                MaxExperimentTimeInSeconds = ExperimentTime,
+                OptimizingMetric = RankingMetric.Ndcg
+            };
+
+            ExperimentResult<RankingMetrics> experimentResult = mlContext.Auto()
+                .CreateRankingExperiment(experimentSettings)
+                .Execute(
+                    trainData: trainDataView,
+                    validationData: testDataView,
```
justinormont, line 93: Should be `validationDataView`:
```suggestion
                    validationData: validationDataView,
```

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -89,6 +98,11 @@ internal static void PrintIterationMetrics(int iteration, string trainerName, Re
             CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.RSquared ?? double.NaN,8:F4} {metrics?.MeanAbsoluteError ?? double.NaN,13:F2} {metrics?.MeanSquaredError ?? double.NaN,12:F2} {metrics?.RootMeanSquaredError ?? double.NaN,8:F2} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
+        internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
+        {
+            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[1] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[0] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
```
justinormont, line 103: Printing NDCG@1,3,10 & DCG@10
```suggestion
            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
```

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -109,6 +123,11 @@ internal static void PrintRegressionMetricsHeader()
             CreateRow($"{"",-4} {"Trainer",-35} {"RSquared",8} {"Absolute-loss",13} {"Squared-loss",12} {"RMS-loss",8} {"Duration",9}", Width);
         }
 
+        internal static void PrintRankingMetricsHeader()
+        {
+            CreateRow($"{"",-4} {"Trainer",-14}, {"nDCG@1",9}, {"nDCG@2",9}, {"nDCG@10",9}, {"DCG@1",9}, {"Duration",9}", Width);
```
justinormont, line 128: Printing NDCG@1,3,10 & DCG@10
```suggestion
            CreateRow($"{"",-4} {"Trainer",-14}, {"nDCG@1",9}, {"nDCG@3",9}, {"nDCG@10",9}, {"DCG@10",9}, {"Duration",9}", Width);
```

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,232 @@
...
+        const string TestDatasetUrl = "https://aka.ms/mlnet-resources/benchmarks/MSLRWeb10KTest240kRows.tsv";
+
+        readonly static string InputPath = Path.Combine(AssetsPath, "Input");
+        readonly static string OutputPath = Path.Combine(AssetsPath, "Output");
+        readonly static string TrainDatasetPath = Path.Combine(InputPath, "MSLRWeb10KTrain720kRows.tsv");
+        readonly static string ValidationDatasetPath = Path.Combine(InputPath, "MSLRWeb10KValidate240kRows.tsv");
+        readonly static string TestDatasetPath = Path.Combine(InputPath, "MSLRWeb10KTest240kRows.tsv");
+        readonly static string ModelPath = Path.Combine(OutputPath, "RankingModel.zip");
+
+        private static uint ExperimentTime = 60;
```
justinormont, line 28: How many models are created in 60 seconds? Likely better w/ longer runtime.

```suggestion
        // Runtime should allow for the sweeping to plateau, which begins near iteration 60
        private static uint ExperimentTime = 600;
```

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
justinormont: Setting to sort on NDCG@10 to match how [AutoML selects the best model](https://github.com/dotnet/machinelearning/blob/f87a3bbd8adb12934dae0a0060813ce9b7500664/src/Microsoft.ML.AutoML/Experiment/MetricsAgents/RankingMetricsAgent.cs#L31-L32):
```suggestion
                .OrderByDescending(r => r.ValidationMetrics.NormalizedDiscountedCumulativeGains[9]).Take(3);
```

We could still match the logic a bit closer (by duplicating the code to here).

justinormont (approved):
LGTM; thanks so much for adding a sample.

Two minor/optional items:
* Matching the sort order, so the sample chooses the same best model as the AutoML code https://github.com/dotnet/machinelearning-samples/pull/852#discussion_r501649673
* Checking on the runtime https://github.com/dotnet/machinelearning-samples/pull/852#pullrequestreview-504683105

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Ranking.csproj
```diff
@@ -0,0 +1,22 @@
...
+	<ItemGroup>
+    <PackageReference Include="Microsoft.ML.AutoML" Version="0.17.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Assets\" />
+    <Folder Include="MLModel\" />
+  </ItemGroup>
+
+</Project>
```
justinormont, line 22: Can also benefit from a README.md for the sample, and link from the repo's main [README.md](https://github.com/dotnet/machinelearning-samples/blob/master/README.md#automl-api-samples-preview-state).

For the sample's README.md, I think the main work would be modifying the [existing ranking readme](https://github.com/dotnet/machinelearning-samples/tree/master/samples/csharp/getting-started/Ranking_Web) by replacing in the AutoML code.

_(tagged on a line of code to allow for a threaded discussion)_

samples/csharp/common/AutoML/ConsoleHelper.cs
jwood803: `NormalizedDiscountedCumulativeGains` and `DiscountedCumulativeGains` only has a length of 3 in the iterations until we get the option to pass in how many of these we want. Should we keep it printing indexes 0, 1, 2, for nDCG and 2 for DCG?

justinormont: Interesting. Where does the limit come from?

justinormont: Looks like the limit come from here:
https://github.com/dotnet/machinelearning/blob/f87a3bbd8adb12934dae0a0060813ce9b7500664/src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs#L34-L38
```c#
        /// <value>
        /// Maximum truncation level for computing (N)DCG
        /// </value>
        [Argument(ArgumentType.AtMostOnce, HelpText = "Maximum truncation level for computing (N)DCG", ShortName = "t")]
        public int DcgTruncationLevel = 3;
```

Few choices:
* Print NDCG/DCG@1,2,3 (as you purpose)
* Change the parameter [default](https://github.com/dotnet/machinelearning/blob/f87a3bbd8adb12934dae0a0060813ce9b7500664/src/Microsoft.ML.Data/Evaluators/RankingEvaluator.cs#L38) from 3 to 10. Should be safe as it simply returns a longer array of DCG/NDCG values.
* Hard code a value to 10 in the AutoML ranking code -- set `var rankingEvaluatorOptions = new RankingEvaluatorOptions { DcgTruncationLevel = 10 };` at https://github.com/dotnet/machinelearning/blob/f87a3bbd8adb12934dae0a0060813ce9b7500664/src/Microsoft.ML.AutoML/Experiment/MetricsAgents/RankingMetricsAgent.cs#L62
* Larger PR to expose the `DcgTruncationLevel` to the AutoML API (as mentioned above: https://github.com/dotnet/machinelearning-samples/pull/852#discussion_r501628536)

jwood803: I just noticed that's where it comes from, as well. 😄

If this isn't in a hurry to get merged in I could make a PR to set the default to 10 and then update this once it gets released. I could also do the reverse to print 1,2, and 3 for now, put the PR in to default to 10, then update this once it gets released.

I'll also create the issue to expose the truncation level to the AutoML API to keep track of the longer term fix.

jwood803: Issue created [here](https://github.com/dotnet/machinelearning/issues/5425) to implement the `DcgTruncationLevel` into the AutoML API.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Ranking.csproj
jwood803: README file was now added. Good suggestion!

jwood803:
The issue to [add the truncation](https://github.com/dotnet/machinelearning/issues/5425) to the AutoML API may be in soon, so we may be able to wait for that and update this accordingly.

samples/csharp/getting-started/Ranking_AutoML/Ranking/README.md
```diff
@@ -0,0 +1,206 @@
...
+* Travel Agency - Provide a list of hotels with those that are most likely to be purchased/booked by the user positioned highest in the list.
+* Shopping - Display items from a product catalog in an order that aligns with a user's shopping preferences.
+* Recruiting - Retrieve job applications ranked according to the candidates that are most qualified for a new job opening.
+
+Ranking is useful to any scenario where it is important to list items in an order that increases the likelihood of a click, purchase, reservation, etc.
+ 
+In this sample, we show how to apply ranking to search engine results. To perform ranking, there are two algorithms currently available - FastTree Boosting (FastRank) and Light Gradient Boosting Machine (LightGBM). We use the LightGBM's LambdaRank implementation in this sample to automatically build an ML model to predict ranking.
```
justinormont, line 13:
```suggestion
In this sample, we show how to apply ranking to search engine results. To perform ranking, there are two algorithms currently available - FastTree Boosting (FastRank) and Light Gradient Boosting Machine (LightGBM). AutoML sweeps over the models and associated hyperparameters to optimize [NDCG@N](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.rankingmetrics.normalizeddiscountedcumulativegains?view=ml-dotnet) (normalized discounted cumulative gains) or [DCG@N](https://docs.microsoft.com/en-us/dotnet/api/microsoft.ml.data.rankingmetrics.discountedcumulativegains?view=ml-dotnet) (discounted cumulative gains).
```

Notation note: I also used `@N` in the text here, as our docs reference `NDCG@N` (in other places, I was using `@k`).

jwood803:
@justinormont 1.5.3 got pushed with the DcgTruncation change so I updated to use it. Just let me know if I missed anything.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,234 @@
...
+            // model will do in production.
+            Console.WriteLine("\n===== Refitting on train+valid and evaluating model's nDCG with test data =====");
+            var trainPlusValidationDataView = textLoader.Load(new MultiFileSource(TrainDatasetPath, ValidationDatasetPath));
+
+            var refitModel = experimentResult.BestRun.Estimator.Fit(trainPlusValidationDataView);
+
+            IDataView predictionsRefitOnTrainPlusValidation = refitModel.Transform(validationDataView);
+
+            // Setting the DCG truncation level
+            var rankingEvaluatorOptions = new RankingEvaluatorOptions { DcgTruncationLevel = 10 };
```
justinormont, line 114: Incase they choose a larger `OptimizationMetricTruncationLevel` above, ensure we produce the requested `@N`:
```suggestion
            var rankingEvaluatorOptions = new RankingEvaluatorOptions
                {
                    DcgTruncationLevel = Math.Min(10, experimentSettings.OptimizationMetricTruncationLevel) 
                };
```

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            Console.WriteLine("===== Download is finished =====\n");
+        }
+
+        private static void PrintTopModels(ExperimentResult<RankingMetrics> experimentResult)
+        {
+            // Get top few runs ranked by nDCG
+            var topRuns = experimentResult.RunDetails
+                .Where(r => r.ValidationMetrics != null && !double.IsNaN(r.ValidationMetrics.NormalizedDiscountedCumulativeGains[0]))
+                .OrderByDescending(r => r.ValidationMetrics.NormalizedDiscountedCumulativeGains[9]).Take(5);
```
justinormont, line 223: I'm concerned users will modify `OptimizationMetricTruncationLevel` above, and being hard hard coded here, will become out of sync. 

The sorting here is different than in other places, which also hard code `[9]`. In other places, it's used for display purposes only and we call out that it's displaying `NDCG@10`. Here, the `experimentResult.BestRun` may not match `topRuns[0]` if the user modifies `OptimizationMetricTruncationLevel`.

Do you see a good way to access, or pass along, the `OptimizationMetricTruncationLevel`? I don't think there's a way to access the `RankingExperimentSettings` from `experimentResult`. So perhaps it's easier to just add a parameter to `PrintTopModels()`:
```diff
- private static void PrintTopModels(ExperimentResult<RankingMetrics> experimentResult)
+ private static void PrintTopModels(ExperimentResult<RankingMetrics> experimentResult, uint optimizationMetricTruncationLevel)
```
<br/>

And of course use here:
```suggestion
                .OrderByDescending(r => r.ValidationMetrics.NormalizedDiscountedCumulativeGains[optimizationMetricTruncationLevel - 1]).Take(5);
```
<br/>

And list the specific `@N`: _(and changing wording from "ranked" to "sorted")_
```diff
- Console.WriteLine("Top models ranked by nDCG --");
+ Console.WriteLine($"Top models sorted by nDCG@{optimizationMetricTruncationLevel} --");
```

justinormont (approved):
LGTM.

Minor items: https://github.com/dotnet/machinelearning-samples/pull/852#discussion_r545794250 & https://github.com/dotnet/machinelearning-samples/pull/852#discussion_r545798552.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,234 @@
...
+            // In the predictions, get the scores of the search results included in the first query (e.g. group).
+            var searchQueries = mlContext.Data.CreateEnumerable<RankingPrediction>(_predictions, reuseRowObject: false);
+            var firstGroupId = searchQueries.First().GroupId;
+            var firstGroupPredictions = searchQueries.Take(100).Where(p => p.GroupId == firstGroupId).OrderByDescending(p => p.Score).ToList();
+
+            // The individual scores themselves are NOT a useful measure of result quality; instead, they are only useful as a relative measure to other scores in the group. 
+            // The scores are used to determine the ranking where a higher score indicates a higher ranking versus another candidate result.
+            foreach (var prediction in firstGroupPredictions)
+            {
+                Console.WriteLine($"GroupId: {prediction.GroupId}, Score: {prediction.Score}");
```
justinormont, line 169: This prints:
```md
=============== Testing prediction engine ===============
#########################################################
=============== Loaded Model OK  ===============
GroupId: 94629043, Score: 10.5748415
GroupId: 94629043, Score: 8.747048
GroupId: 94629043, Score: 8.339484
GroupId: 94629043, Score: 8.295864
GroupId: 94629043, Score: 7.7126718
GroupId: 94629043, Score: 7.094361
GroupId: 94629043, Score: 6.403935
GroupId: 94629043, Score: 5.6126056
GroupId: 94629043, Score: 5.4343157
GroupId: 94629043, Score: 5.341767
GroupId: 94629043, Score: 3.4955919
GroupId: 94629043, Score: 3.0068336
GroupId: 94629043, Score: 2.925596
...
```

Is there more information we can print? Currently the user has no way to compare the ranked results against the original input data.

Could we print the correct/input Label in a column? The user could then verify the input Label value is decreasing in the list.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,234 @@
...
+        readonly static string OutputPath = Path.Combine(AssetsPath, "Output");
+        readonly static string TrainDatasetPath = Path.Combine(InputPath, "MSLRWeb10KTrain720kRows.tsv");
+        readonly static string ValidationDatasetPath = Path.Combine(InputPath, "MSLRWeb10KValidate240kRows.tsv");
+        readonly static string TestDatasetPath = Path.Combine(InputPath, "MSLRWeb10KTest240kRows.tsv");
+        readonly static string ModelPath = Path.Combine(OutputPath, "RankingModel.zip");
+
+        // Runtime should allow for the sweeping to plateau, which begins near iteration 60
+        private static uint ExperimentTime = 600;
+
+        private static IDataView _predictions = null;
```
justinormont, line 31: Could `_predictions` be passed around instead of at this scope?

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -109,6 +123,11 @@ internal static void PrintRegressionMetricsHeader()
             CreateRow($"{"",-4} {"Trainer",-35} {"RSquared",8} {"Absolute-loss",13} {"Squared-loss",12} {"RMS-loss",8} {"Duration",9}", Width);
         }
 
+        internal static void PrintRankingMetricsHeader()
+        {
+            CreateRow($"{"",-4} {"Trainer",-14}, {"nDCG@1",9}, {"nDCG@3",9}, {"nDCG@10",9}, {"DCG@10",9}, {"Duration",9}", Width);
```
justinormont, line 128: Extra commas:
```suggestion
            CreateRow($"{"",-4} {"Trainer",-14} {"nDCG@1",9} {"nDCG@3",9} {"nDCG@10",9} {"DCG@10",9} {"Duration",9}", Width);
```

Current prints as:
image

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -89,6 +98,11 @@ internal static void PrintIterationMetrics(int iteration, string trainerName, Re
             CreateRow($"{iteration,-4} {trainerName,-35} {metrics?.RSquared ?? double.NaN,8:F4} {metrics?.MeanAbsoluteError ?? double.NaN,13:F2} {metrics?.MeanSquaredError ?? double.NaN,12:F2} {metrics?.RootMeanSquaredError ?? double.NaN,8:F2} {runtimeInSeconds.Value,9:F1}", Width);
         }
 
+        internal static void PrintIterationMetrics(int iteration, string trainerName, RankingMetrics metrics, double? runtimeInSeconds)
+        {
+            CreateRow($"{iteration,-4} {trainerName,-9} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4}, {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
```
justinormont, line 103: Extra commas & spacing:
```suggestion
            CreateRow($"{iteration,-4} {trainerName,-14} {metrics?.NormalizedDiscountedCumulativeGains[0] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[2] ?? double.NaN,9:F4} {metrics?.NormalizedDiscountedCumulativeGains[9] ?? double.NaN,9:F4} {metrics?.DiscountedCumulativeGains[9] ?? double.NaN,9:F4} {runtimeInSeconds.Value,9:F1}", Width);
```

samples/csharp/common/AutoML/ConsoleHelper.cs
```diff
@@ -55,6 +55,15 @@ public static void PrintMulticlassClassificationMetrics(string name, MulticlassC
             Console.WriteLine($"************************************************************");
         }
 
+        public static void PrintRankingMetrics(string name, RankingMetrics metrics)
+        {
+            Console.WriteLine($"************************************************************");
+            Console.WriteLine($"*    Metrics for {name} ranking model   ");
+            Console.WriteLine($"*-----------------------------------------------------------");
+            Console.WriteLine($"    Discounted Cumulative Gain (DCG@10) = {metrics?.DiscountedCumulativeGains?[9] ?? double.NaN:0.####}");
+            Console.WriteLine($"    Normalized Discounted Cumulative Gain (NDCG@10) = {metrics?.NormalizedDiscountedCumulativeGains?[9] ?? double.NaN:0.####}, a value between 0 and 1, the closer to 1, the better");
```
justinormont, line 64: Wording:
```suggestion
            Console.WriteLine($"    Normalized Discounted Cumulative Gain (NDCG@10) = {metrics?.NormalizedDiscountedCumulativeGains?[9] ?? double.NaN:0.####}, a value from 0 and 1, where closer to 1.0 is better");
```

justinormont: #Resolved

samples/csharp/common/AutoML/ConsoleHelper.cs
justinormont: #Resolved

samples/csharp/common/AutoML/ConsoleHelper.cs
justinormont: #Resolved

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
justinormont: #Resolved

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
justinormont: #Resolved

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
justinormont: #Resolved

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
justinormont: #Resolved

samples/csharp/common/AutoML/ConsoleHelper.cs
justinormont: #Resolved

justinormont:
@jwood803: I checked-in various fixes.

Most fixes were addressing my own review feedback. I also modified the thrice training to use the correct scoring dataset, and to return the refit model.

Only remaining feedback is https://github.com/dotnet/machinelearning-samples/pull/852#pullrequestreview-555497443.

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
```diff
@@ -0,0 +1,221 @@
...
+            {
+                DcgTruncationLevel = Math.Min(10, (int)experimentSettings.OptimizationMetricTruncationLevel * 2)
+            };
+
+            Console.WriteLine("\n===== Evaluating model's NDCG (on test data) =====");
+            IDataView predictions = experimentResult.BestRun.Model.Transform(testDataView);
+            var metrics = mlContext.Ranking.Evaluate(predictions, rankingEvaluatorOptions);
+            ConsoleHelper.PrintRankingMetrics(experimentResult.BestRun.TrainerName, metrics, experimentSettings.OptimizationMetricTruncationLevel);
+
+            // STEP 5: Refit the model with all available data
```
justinormont, line 110:
```suggestion
            // STEP 5: Refit the model and get final metrics
```

samples/csharp/getting-started/Ranking_AutoML/Ranking/Ranking/Program.cs
jwood803: I can add the original label to the output. Would that help?
image

justinormont: Is there a way to do a late merge, without adding `Label` to the `RankingPrediction` class? 

Do you think if we add the `Label` to the output class, some users will misconstrue it as the predicted label (`Score` is the actual prediction)? Seeing no mispredictions, users could assume their model is perfect, and launch it to production only to find the `Label` field is then empty.

Even without the added label output, this PR is very good. I'm ok merging as is, or with adding the correct/input label.

jwood803: My only thought is to do something like this:

```csharp
var predictionsPreview = _predictions.Preview();

for (int i = 0; i < firstGroupPredictions.Count; i++)
{
  var currentPredictionPreview = predictionsPreview.RowView[i].Values;

  Console.WriteLine($"GroupId: {firstGroupPredictions[i].GroupId}, Score: {firstGroupPredictions[i].Score}");
}
```

Use the predictions from the test set and get the label from the `Preview` of it.

I'm not sure if it's the best thing to do or not, though. If this isn't good, perhaps we can get this in and can do another update once we can find a better solution.

justinormont: How does this look:
```c#
// Label values from the test dataset (not the predicted scores/labels)
IEnumerator<string> labelEnumerator = mlContext.Data.CreateEnumerable<RankingData>(testDataView, true).Select(a => a.Label).GetEnumerator();

foreach (var prediction in firstGroupPredictions)
{
    labelEnumerator.MoveNext();
    Console.WriteLine($"GroupId: {prediction.GroupId}, Score: {prediction.Score}, Correct Label: {labelEnumerator.Current}");
}
```

It would require passing in `testDataView`. Technically, the same enumerator can be made from `firstGroupPredictions`, though reading from the test dataset reenforces that the printed label is not part of the prediction, but instead a given label.

jwood803: That works great! Pushed the latest changes with this in it. Thanks!

justinormont: LGTM. Good to merge?

jwood803: I think so! Thank you for helping me with this sample!

PR #854, opened 2020-10-02
Update README.md

Ravikc:
change class name


luisquintanilla (approved):
Thanks for this update @Ravikc

PR #859, opened 2020-11-17
Update README.md

danespinosa:
Typo, I believe it should be price rather than prise.


luisquintanilla (approved):
Hi @danespinosa,

Great catch! Thanks for making this change.

PR #869, opened 2020-12-11
Run explainability winforms fix

luisquintanilla:


luisquintanilla:
@antoniovs1029 This is the PR that should make the winforms app runnable.

PR #871, opened 2020-12-11
Fix TensorFlow samples for .NET Core 2.1 & minor fix in one sample

mstfbl:
Added `System.Runtime.CompilerServices.Unsafe` for .NET Core 2.1 support. 

Also added the 11 MB `flower_photo_small_set` to the `DeepLearning_TensorFlowEstimator` sample as the link to download this is broken, and that this zip has already been added in the `DeepLearning_ImageClassification_Training`. This way, the `DeepLearning_TensorFlowEstimator` sample now works properly.


frank-dong-ms (approved):
:shipit:

PR #882, opened 2021-03-12
add sample for model builder object detection tutorial

briacht:


samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            // Load model & create prediction engine
+            ITransformer mlModel = mlContext.Model.Load(MLNetModelPath, out var modelInputSchema);
+
+            var predEngine = mlContext.Model.CreatePredictionEngine<ModelInput, ModelOutput>(mlModel);
+
+             return predEngine;
+        }
+
+        private static void DownloadModel()
```
ericstj, line 52: Maybe we can move this into the project file instead using https://docs.microsoft.com/en-us/visualstudio/msbuild/downloadfile-task?view=vs-2019.

```
  <ItemGroup>
    <ModelFileUrl Include="https://mlpublicassets.blob.core.windows.net/samples/object-detection/MLModel.zip" />
  </ItemGroup>

  <Target Name="DownloadModelFiles" BeforeTargets="Build">
    <DownloadFile SourceUrl="@(ModelFileUrl)" SkipUnchangedFiles="true" DestinationFolder="$(OutDir)"/>
  </Target>
```

That way you don't need to do this at runtime and it's closer to what we would do if we could check in the model.

ericstj: Also, we might want to use an aka.ms link in case we need to change the storage location in the future.

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.ConsoleApp/StopSignDetectionML.ConsoleApp.csproj
```diff
@@ -0,0 +1,26 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\StopSignDetectionML.Model\StopSignDetectionML.Model.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectCapability Include="ModelBuilderGenerated" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="test-image1.jpeg">
```
ericstj, line 22: Is the JPEG intentionally empty?

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/StopSignDetectionML.Model.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML" Version="1.5.0" />
+    <PackageReference Include="Microsoft.ML.OnnxTransformer" Version="1.5.0" />
+    <PackageReference Include="Microsoft.ML.OnnxRuntime" Version="1.3.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="MLModel.zip">
```
ericstj, line 13: I think this should be removed if we use the download mechanism.

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        // For more info on consuming ML.NET models, visit https://aka.ms/mlnet-consume
+        // Method for consuming model in your app
+        public static ModelOutput Predict(ModelInput input)
+        {
+            ModelOutput result = PredictionEngine.Value.Predict(input);
+            return result;
+        }
+
+        public static PredictionEngine<ModelInput, ModelOutput> CreatePredictionEngine()
```
ericstj, line 33: What's our recommended best practice around `PredictionEngine.Dispose: one static and reuse it or tie it to some other instance and dispose?  It's a heavy object so I can see why we wouldn't want to create one each time we use it, but I get anxious when I see Disposable objects being left for finalization to cleanup.

Youssef1313 (commented):
Would you be able to address https://github.com/dotnet/docs/pull/23300/files#r593655483?

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
briacht: Fixed with https://github.com/dotnet/machinelearning-samples/pull/882/commits/d559ca315873ee6fad927f449f0a690e3d562f8b

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/StopSignDetectionML.Model.csproj
briacht: Fixed with https://github.com/dotnet/machinelearning-samples/pull/882/commits/d559ca315873ee6fad927f449f0a690e3d562f8b

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.ConsoleApp/StopSignDetectionML.ConsoleApp.csproj
briacht: Not sure about that. I added a jpeg as a test image. Is it ok since it's just 1 image if it's part of repo?

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetection/StopSignDetection.csproj
```diff
@@ -13,6 +13,14 @@
     <ProjectReference Include="..\StopSignDetectionML.Model\StopSignDetectionML.Model.csproj" />
   </ItemGroup>
 
+  <Target Name="DownloadModelFiles" BeforeTargets="Build">
+    <DownloadFile 
+      SourceUrl="https://aka.ms/mlnet-object-detection-tutorial-model"
```
ericstj, line 18: Neat, it chooses the right file name for this?

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
```diff
@@ -53,7 +47,7 @@ private static void DownloadModel()
         {
```
ericstj, line 47: Can you delete the code that downloads the model here?

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetection/StopSignDetection.csproj
briacht: Yep! At least it built and ran for me when I tested it

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.ConsoleApp/StopSignDetectionML.ConsoleApp.csproj
ericstj: Should be fine, but for some reason github was showing it as a zero-byte file in the diff.

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetection/StopSignDetection.csproj
michaelgsharp: That aka.ms links directly to MLModel.zip. I believe as long as that is the model file name it will work fine.

ericstj: This was mostly for my info, I haven't used DownloadFile task a lot.  Looks nifty for this sort of thing in a sample.

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.ConsoleApp/ModelBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.ML;
+using Microsoft.ML.Data;
+using StopSignDetectionML.Model;
+namespace StopSignDetectionML.ConsoleApp
+{
+    public static class ModelBuilder
+    {
+        private static string TRAIN_DATA_FILEPATH = @"C:\Users\brachtma\AppData\Local\Temp\fe343e6e-b4bb-4eae-94e3-ade20e72b9a0.csv";
```
michaelgsharp, line 13: This file path isn't going to work correctly.

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.ConsoleApp/StopSignDetectionML.ConsoleApp.csproj
ericstj: Nevermind, if I click "view" I can see it.

michaelgsharp: I saw that too, but if you actually click on view the file it shows the actual picture. Seems github is just displaying that wrong.

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
briacht: Whoops yes! Fixed with https://github.com/dotnet/machinelearning-samples/pull/882/commits/0a55fa5ebcb0c763c125bc17c511d57f56f59f9a

briacht:
> Would you be able to address https://github.com/dotnet/docs/pull/23300/files#r593655483?

Done

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Web;
+using System.Net;
+
+namespace StopSignDetectionML.Model
+{
+    public class ConsumeModel
+    {
+        private static Lazy<PredictionEngine<ModelInput, ModelOutput>> PredictionEngine = new Lazy<PredictionEngine<ModelInput, ModelOutput>>(CreatePredictionEngine);
+
+        private static WebClient client = new WebClient();
```
ericstj, line 21: nit: this is now unused

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
michaelgsharp: The prediction engine isn't thread safe. So both things you suggested have been done. If we can reuse it we should, but if its in a threaded environment we can't.

michaelgsharp (changes requested):
At least the filepath needs to be changed before this can go in.

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.ConsoleApp/ModelBuilder.cs
briacht: I meant to delete the console app. I removed from Solution Explorer but didn't delete I guess.

briacht: Fixed with https://github.com/dotnet/machinelearning-samples/pull/882/commits/5b830114cf7e91b2b9f294aa6f38240e582d6e57

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
briacht: Fixed with https://github.com/dotnet/machinelearning-samples/pull/882/commits/63c54104961e38d70c93e59db6da9687feb9b568

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetectionML.Model/ConsumeModel.cs
ericstj: > both things you suggested have been done

I'm not asking about PredictionEngine itself, but how it is used.

If it's not thread safe then our sample code shouldn't be wrapping it in a static.  Maybe since there's not much happening in this class any more we could move CreatePredictionEngine into Program.cs, then just do this:
```c#
using (var predictionEngine = CreatePredictionEngine())
{
    var predictionResult = ConsumeModel.Predict(sampleData);
}
```

Would that still achieve the goal of the sample?

Alternatively, if we're trying demonstrate thread-safe prediction we could keep the static class and use a ThreadLocal for the engine.

briacht: Not sure about this, but this is the code that's generated by Model Builder

briacht: This is the code generated by Model Builder, so if there's issues with PredictionEngine then might have to work with tooling team on changing code generation as well

michaelgsharp: If its generated by model builder does it need to be checked in? Or did you just want an example of what it will look like?

briacht: I still need the sample to be able to run (users can download the whole sample or follow the tutorial step by step)

michaelgsharp: So I think that looking into changing the code generated by the model builder would be good since @ericstj is right that we shouldn't have non-thread safe code be static, but since the code is auto generated and not anything you have modified I don't think we should block that here.

@ericstj thoughts?

ericstj: I hadn't realized this was auto-generated.  If that's the case, let's revert the code to the exact version that's auto-generated and treat the characteristics of the auto-generated code as a separate issue.  cc @eerhardt @JakeRadMSFT

Youssef1313:
> > Would you be able to address https://github.com/dotnet/docs/pull/23300/files#r593655483?
> 
> Done

I can't see this added in samples browser with any commit. Have you forgot to push this change?

briacht:
@michaelgsharp @ericstj  So can this be merged?

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetection/README.txt
```diff
@@ -1 +1,40 @@
-﻿Test image is from pexels.com: https://www.pexels.com/photo/red-stop-sign-39080/
\ No newline at end of file
+﻿---
```
Youssef1313, line 1: extension should me `md`

briacht: https://github.com/dotnet/machinelearning-samples/pull/882/commits/fb73c03cbd2ec1e4ca0f747c0e7c6dbe462498c7

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetection/README.txt
```diff
@@ -1 +1,40 @@
...
+page_type: sample
+name: "ML.NET Model Builder Stop Sign Object Detection"
+description: "Use ML.NET Model Builder to train a deep learning object detection model in Azure."
+urlFragment: "mlnet-object-detection-model-builder"
+languages:
+- csharp
+products:
+- dotnet
+- dotnet-core
+- ml-dotnet
```
Youssef1313: I'm not sure if `ml-dotnet` is a valid value here.

In <https://docs.microsoft.com/samples/browse/>, the products shown at the left side doesn't contain ML. You may want to have an internal ask for a node under .NET for ML?

briacht: Removed ml-dotnet and added visual-studio in latest commit

samples/modelbuilder/ObjectDetection_StopSigns/StopSignDetection/README.md
```diff
@@ -0,0 +1,40 @@
+﻿---
```
Youssef1313, line 1: There is an invisible BOM character here that seems to prevent this from rendering correctly.

Before:

image


After:

image

This may prevent it from being recognized by @VSC-Service-Account

```suggestion
---
```

briacht: It doesn't let me commit your suggestion since there's no recognizable changes.

Does the latest commit fix it?

Youssef1313: @briacht Nope. https://github.com/briacht/machinelearning-samples/pull/1 should do it.

michaelgsharp: Once his PR is merged into your code @briacht I'll go ahead and approve.

michaelgsharp (approved):
LGTM.

ericstj (commented):
LGTM. Thanks for the sample!

PR #891, opened 2021-04-01
Add onnx export (regression) sample

briacht:


samples/csharp/getting-started/ONNXExport/ONNXExport/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+| ML.NET version | API type          | Status                        | App Type    | Data type | Scenario            | ML Task                   | Algorithms                  |
+|----------------|-------------------|-------------------------------|-------------|-----------|---------------------|---------------------------|-----------------------------|
+| v1.5.5           | Dynamic API | Up-to-date | Console app | .csv files | Price prediction | Regression  | Light GBM regression |
+
+In this sample, you'll see how to use ML.NET to train a regression model and then convert that model to the ONNX format.
+
+## Problem
+
+The Open Neural Network eXchange i.e [ONNX](http://onnx.ai/) is an open format to represent deep learning models. With ONNX, developers can move models between state-of-the-art tools and choose the combination that is best for them. ONNX is developed and supported by a community of partners.
```
michaelgsharp, line 11: I know the X is capitalized so the letters spell ONNX, but their website still capitalizes it as `Open Neural Network Exchange` so I would match that.

samples/csharp/getting-started/ONNXExport/ONNXExport/ONNXExport/Program.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            var trainingPipeline = dataProcessPipeline.Append(trainer);
+
+            // Train ML.NET model on training data
+            ITransformer model = trainingPipeline.Fit(trainingDataView);
+
+            // You need a transformer and input data to convert an ML.NET model to an ONNX model
+            // By default, the ONNX conversion will generate the ONNX file with the latest OpSet version
+            using (var stream = File.Create("taxi-fare-model.onnx"))
+                mlContext.Model.ConvertToOnnx(model, trainingDataView, stream);
```
michaelgsharp, line 49: Do you think we should talk about how they can pick which columns to output from the ONNX model? Or is that more advanced than this tutorial should cover?

samples/csharp/getting-started/ONNXExport/ONNXExport/README.md
briacht: I copy + pasted this from another of the ONNX samples so probably best to fix it there as well! I'll see if I can find it again and fix.

samples/csharp/getting-started/ONNXExport/ONNXExport/ONNXExport/Program.cs
briacht: What would be the use case for this and is there an example I can look at to bring over?

michaelgsharp: So we added it in in [1.5.4](https://github.com/dotnet/machinelearning/releases/tag/v1.5.4). Its gives the user more performance. By default, since we don't know which columns the user wants, we are outputting all of the columns from the onnx model. This works great, but makes the model bigger and decreases performance. For many situations that wont matter, but in cases where they do need good performance it can make a very large difference.

luisquintanilla (approved):
LGTM

PR #922, opened 2021-08-25
Update Anomaly Detection Sample to Latest Version

jwood803:
Updating the Anomaly Detection - Phone Calls sample to the latest ML.NET version

cc// @luisquintanilla


luisquintanilla:
Hold until dotnet/machinelearning#5891 is investigated / fixed.

PR #938, opened 2021-12-17
Update .vsts-dotnet-ci.yml for Azure Pipelines

luisquintanilla:


luisquintanilla:
@michaelgsharp can I get an approval when you get a chance please. Thanks!

michaelgsharp (approved):
We can probably refactor this further using templates, but since its not very complex we can get to that if we need it in the future.

PR #986, opened 2023-03-22
Update ObjectDetection-Onnx README.md

DavideDunne:
Let the user know he can use Visual Studio 2022 or subsequent versions.


DavideDunne:
@dotnet-policy-service agree

PR #993, opened 2023-05-17
Update README.md

alexiftode:
Not a big issue but I was a bit confused if there is a missing file. I believe HeartTraining.tsv should be HeartTraining.csv to reflect the real present files.


alexiftode:
@dotnet-policy-service agree

SeanEngineering (approved):
Also agree with this, may have been following naming convention of other .tsv files

PR #994, opened 2023-06-27
Clean up metadata

gewarren:
We're trying to remove the dotnet-core product label for the site experience, and that's not the main technology for these samples, so I removed it (also removed the VS products since they're used in the samples, but they're not what the sample is *about*).

Also other whitespace and spelling cleanup.

Contributes to https://github.com/dotnet/docs/issues/35970


gewarren:
@luisquintanilla Can you review?

luisquintanilla (approved):
LGTM. Thank you @gewarren

PR #1017, opened 2024-01-17
Updated samples image url

michaelgsharp:
Updated the flower photos sample dataset url to the correct url.


ericstj (approved):
Thanks for the cleanup!

